PR #49, opened 2018-01-05
start

anri-1151:


app/src/angular/controllers/commons/dialogs/send-token-controller.js
```diff
@@ -15,13 +15,15 @@ function SendTokenDialogController($rootScope, $scope, $log, $q, $mdDialog, $int
...
+    $scope.gasLimit = args.gasLimit || 210000;
 
     $scope.data = {
-        sendAmount: null,
-        sendToAddress: '',
+        sendAmount: args.sendAmount || null,
+        sendToAddress: args.sendToAddress || '',
         txStatus: null
     }
+    $scope.editModeOfAddress = args.editModeOfAddress;
```
and1gio: set *true* if arg is missing

app/src/angular/controllers/commons/dialogs/send-token-controller.js
```diff
@@ -15,13 +15,15 @@ function SendTokenDialogController($rootScope, $scope, $log, $q, $mdDialog, $int
...
+    $scope.editModeOfPrice = args.editModeOfPrice;
```
and1gio: same here.. set *true* if arg is missing

and1gio:
@anri-1151 
also instead of lock **send amount** you should lock **gas price**

PR #65, opened 2018-01-13
Wallet eth activity

tmchedlishvili:


app/src/angular/classes/token.js
```diff
@@ -104,6 +104,12 @@ class Token {
 
             if(balanceHex !== oldBalanceHex){
                 $rootScope.$broadcast('balance:change', this.symbol, this.getBalanceDecimal(), this.balanceInUsd);
+                if (this.symbol.toUpperCase() == 'ETH') {
```
and1gio: this will never be "ETH" they all are ERC-20 Tokens

PR #74, opened 2018-01-19
IDWAL00K-220: Merge IDWAL00K-220 to 1/14

jvsam:


anri-1151:
back -ზე რომ აწვები და უკან გაბრუნებს მემგონო სწორ ფორმაზე დავაბრუნე. თუ რამეა ჩავასწორებ უცბად.

app/src/templates/guest/welcome.pug
```diff
@@ -3,7 +3,7 @@
     .content(flex, layout="column", layout-align="center center") 
         .logo
         .header selfkey
-        .subheader blockchain identity wallet
+        .subheader blockchain identity wallet member.setup.choose
```
and1gio: why there is added **member.setup.choose** at the end of this line ?

anri-1151: It's nothing. I will delete it. Maybe I copy it and paste it by accident.

anri-1151:
Changes are made only in "app/src/templates/member/setup/choose.pug". Other file changes are made by IDE when I opened them. It is only spaces and etc. nothing else. And it's already pushed and from it these files can't be deleted. So you can merge it. This will not cause problems.

PR #83, opened 2018-01-25
dock-icon

burdula:


and1gio:
there are lots of changes.. don't reformat file

PR #98, opened 2018-02-13
info

burdula:


app/src/templates/common/dialogs/info-dialog.pug
```diff
@@ -0,0 +1,7 @@
+md-dialog.info(aira-lable="Dialog Title")
```
and1gio: pass dialog title as param

app/src/templates/common/dialogs/info-dialog.pug
```diff
@@ -0,0 +1,7 @@
...
+    .info-wrapper
+        .info-text
+            span(ng-if="form !== 'version'") Coming Soon
```
and1gio: pass text as param

and1gio: remove ng-if .. just interpolate the text param

app/src/angular/controllers/member/right-sidenav-controller.js
```diff
@@ -20,6 +20,23 @@ function MemberRightSidenavController($rootScope, $scope, $log, $mdSidenav, $sta
         $scope.close();
     }
 
+
+    $rootScope.sidenavInfoDialog = (event, item) => {
```
and1gio: move in app.run.js & rename to *openInfoDialog*

PR #101, opened 2018-02-17
Idwal00 s 15

anri-1151:
Task is finished as it it is described in jira. The CMC data is saved in main-store.json as cmcData = {prices: [] } and each item has image path on cmc web site and also there are icons saved in local-machine in cmc-icos folder. Update tame and image sizes are cinfigured from config.js


anri-1151:
New method is added to load obliged image icons, which is possible to call from front service.

PR #127, opened 2018-03-19
fix

anri-1151:
I have fixed these issued. Default value must be choose a token in select box and when change it address should be stay the same.


and1gio:
@anri-1151 pls test the ETH Send dialog.. this commit brakes it

PR #128, opened 2018-03-20
pie chart visibility bug fix

tmchedlishvili:


app/src/templates/member/dashboard/main.pug
```diff
@@ -23,7 +23,7 @@
                         | <div style="color: #ffffff; font-size:14px; font-weight: bold;"> You've reached the limit and cannot add another token.
                     sk-icon(icon="plus", size-class="icon-size-26x26")
             .sk-horizontal-devider
-            sk-circle-pie-chart(data="pieChart", ng-if="totalBalanceInUsd > 0")
+            sk-circle-pie-chart(data="pieChart", ng-if="wallet.calculateTotalBalanceInUSD() > 0")
```
and1gio: right now i'm accepting this.. but also think about change it.. and also i'll link other places where are used same technic.. we need to change that parts also..

PR #133, opened 2018-03-23
Idwal00 s 71

gmerabishvili:


app/src/templates/member/marketplace/exchange-item.pug
```diff
@@ -8,15 +8,16 @@
                     img.logo-image(ng-src="{{realData.logo[0].thumbnails.small.url}}", alt="")
                     span.name {{realData.name}}
                 .status
-                    md-icon(md-svg-src="assets/images/icons/ic_help_outline_black_24px.svg" class="color-white")
                     div.text Your Staking Status
+                        md-icon(md-svg-src="assets/images/icons/ic_help_outline_black_24px.svg" class="color-white")
                             md-tooltip.sk-tooltip(md-direction="bottom")
                                 |  <div style="text-align:left;"> What is Staking? <br />
```
and1gio: @gmerabishvili will this work? there is embed html inside pug .. i think compiler will throw errors..

gmerabishvili: yes, it will work

PR #136, opened 2018-03-24
Idwal00 s 69

burdula:


and1gio:
@burdula fix conflicts and also .. don't commit main.css file

PR #169, opened 2018-04-10
sel-19

burdula:


app/src/angular/directives/commons/keyPress.directive.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function keyPressDirective($document) {
```
and1gio: pass target KEY as argument to this directive.. by default use 13 === enter

app/src/angular/controllers/commons/dialogs/id-attributes/add-edit-static-data-controller.js
```diff
@@ -54,6 +54,8 @@ function AddEditStaticDataDialogController($rootScope, $scope, $log, $q, $mdDial
 
 
     $scope.save = (event, theForm) => {
+
+        console.log(777,theForm)
```
and1gio: remove log

app/src/angular/directives/commons/keyPress.directive.js
and1gio: rename directive.. to: onKeyPressDirective.. 
```
<button on-key-press="someFunction($event, item)" on-key="13">Hit Enter</button>
```

PR #176, opened 2018-04-19
Sel 18

burdula:


app/src/angular/services/common.service.js
```diff
@@ -63,16 +63,21 @@ function CommonService($rootScope, $log, $q, $mdDialog, $compile, $mdToast) {
             return result;
         }
 
-        commasAfterNumber (amount, num) {
+        commasAfterNumber(amount, num) {
```
dfcowell: Why are there two declarations of the same function name in one file? See line 48.

app/src/angular/controllers/member/wallet/manage-token-controller.js
```diff
@@ -31,19 +31,24 @@ function ManageTokenController($rootScope, $scope, $state, $log, $mdDialog, $sta
     function prepareBalance() {
         if ($scope.symbol === 'ETH') {
             // ETHER
-            $scope.balance = CommonService.commasAfterNumber($rootScope.wallet.balanceEth, 2);
-            $scope.balanceUsd = CommonService.commasAfterNumber($rootScope.wallet.balanceInUsd, 2);
+            let ethRoundedBalanceUSD = $rootScope.wallet.balanceInUsd > 0 ? $rootScope.wallet.balanceInUsd.toFixed(2) : $rootScope.wallet.balanceInUsd;
```
dfcowell: Why aren't we using the [built-in formatting functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat) that do all this for us?

burdula: we have a service in CommonService which puts commas after numbers we want to. we can use this built-in function instead that, but we still have to use toFixed method, because we want to show only two decimals rounded in the end, when its not whole number.

dfcowell: >but we still have to use toFixed method, because we want to show only two decimals rounded in the end, when its not whole number.

No you don't. The built-in function automatically rounds to the correct number of digits as long as you specify `{ style: 'currency', currency: 'USD' }` in the options.

app/src/angular/services/common.service.js
burdula: don't know who worked(used) on that(first one),maybe some merge problem.I commented first one in case someone still needs that.

PR #181, opened 2018-04-24
Fix sel 17

and1gio:


and1gio:
@altninja unfortunately nop.. md-select is a directive that creates some additional dom elements.. and styles are applied to that elements..

dfcowell:
@altninja what's the status of this PR?

PR #186, opened 2018-05-07
feat(sqlite): DB migrations for SQLite using knex

altninja:
auto migrate on app start based on knex migration files - can use `npm run mig-new -- migration_name`
to add a new migrations file and `npm run mig-test` to run a test


app/controllers/sql-lite-service.js
```diff
@@ -8,18 +8,30 @@ const countriesList = require('./../../assets/data/country-list.json');
...
+        const log = ({ action, migration }) => console.log('Doing ' + action + ' on ' + migration)
+        try {
+            await knexMigrate('up', { to: mv}, log)
+            await knexMigrate('down', { to: mv}, log)
+        } catch (e) {
+            console.log(e)
         }
-    });
+    }
+    migrations()
```
dfcowell: This is non-blocking and could cause problems due to a race condition later on if there are required migrations that the application depends upon during start-up.

Putting it in an IIFE and using async-await to block on it would solve that problem.

migrations/20180507145702_test1.js
```diff
@@ -0,0 +1,15 @@
+
+exports.up = function(knex, Promise) {
```
dfcowell: This whole file is going to break many things.

migrations/20180507170725_test2.js
```diff
@@ -0,0 +1,15 @@
+
+exports.up = function(knex, Promise) {
+   return Promise.all([
+    knex.schema.createTable('check2', function(table){
```
dfcowell: We definitely don't need this one either.

knexfile.js
```diff
@@ -0,0 +1,26 @@
...
+const OSENV = process.env.OSENV || 'osx'
+
+let userDataPath
+
+if (OSENV === 'windows') {
+	userDataPath = '%APPDATA%\\Electron\\'
+} else if (OSENV === 'linux') {
+	userDataPath = '~/.config/Electron/'
+} else {
+	userDataPath = '/Users/' + user + '/Library/Application Support/Electron/'
```
dfcowell: Prefer template strings (using backticks and `${user}` syntax)

test/migrate.js
```diff
@@ -0,0 +1,15 @@
+const 
+	knexFile = require('../knexfile.js')
+	knex = require('knex')(knexFile)
+	knexMigrate = require('knex-migrate')
+	mv = '20180507173236'
+
+async function run() {
+	const log = ({ action, migration }) => console.log('Doing ' + action + ' on ' + migration)
```
dfcowell: Prefer template strings

app/controllers/sql-lite-service.js
```diff
@@ -135,24 +136,24 @@ module.exports = function (app) {
...
-        promises.push(IdAttribute.init());
-        promises.push(TokenPrice.init());
-        promises.push(createWalletTokens());
-        promises.push(TransactionHistory.init());
-        promises.push(ActionLog.init());
-        promises.push(WalletSetting.init());
-        promises.push(ExchangeDataHandler.init());
-        return Promise.all(promises)
-    }
+    // controller.prototype.init = () => {
```
dfcowell: This can be deleted.

app/controllers/sql-lite-service.js
```diff
@@ -75,30 +76,30 @@ module.exports = function (app) {
...
-                    }).catch((error) => {
-                        reject(error);
-                    });
-                } else {
-                    resolve();
-                }
-            });
-        });
-    }
+    // function createWalletTokens() {
```
dfcowell: Can this be deleted?

altninja: yeah it is replicated in the seed file along with the others

app/controllers/sql-lite-service.js
altninja: thanks, will delete all commented code next commit

test/migrate.js
altninja: sure can do, let me look for a better syntax highlighting solution, its hard to read on longer strings without the concats showing where the vars are

app/controllers/sql-lite-service.js
altninja: yes can delete this, its been replicated in the seed file now

test/migrate.js
dfcowell: If you're not using VSCode you're doing it wrong. 😛

altninja:
closing this, merged into standards PR

PR #187, opened 2018-05-09
Sel 13

tmchedlishvili:


app/controllers/models/wallet.js
```diff
@@ -40,6 +40,7 @@ module.exports = function (app, sqlLiteService) {
                         table.string('keystoreFilePath');
                         table.binary('profilePicture');
                         table.integer('isSetupFinished').notNullable().defaultTo(0);
+                        table.integer('isLedger').notNullable().defaultTo(0);
```
dfcowell: This should be a string or enum field specifying a `provider` or `profile` which can be any of:

- local (private key stored in wallet file)
- ledger
- trezor (not implemented yet)

app/src/angular/classes/wallet.js
```diff
@@ -254,11 +257,21 @@ class Wallet {
             if (contractDataHex) {
                 rawTx.data = EthUtils.sanitizeHex(contractDataHex);
             }
-
-            let eTx = new Tx(rawTx);
-            eTx.sign(this.privateKey);
-
-            defer.resolve('0x' + eTx.serialize().toString('hex'));
+            
+            if (!this.isLedger) {
```
dfcowell: We should be using the factory pattern here to allow for adding other integrations in the future.

app/src/angular/configs/app.run.js
```diff
@@ -36,13 +36,15 @@ function AppRun($rootScope, $log, $window, $timeout, $interval, $q, $state, $tra
     Wallet.SqlLiteService = SqlLiteService;
     Wallet.CommonService = CommonService;
     Wallet.EtherScanService = EtherScanService;
+    Wallet.LedgerService = LedgerService;
```
dfcowell: Why are we monkeypatching these modules like this?

app/src/angular/controllers/commons/dialogs/send-token-controller.js
```diff
@@ -192,13 +201,18 @@ function SendTokenDialogController($rootScope, $scope, $log, $q, $mdDialog, $int
...
+                    CommonService.showToast('error', error, 20000);
+                    $scope.errors.sendFailed = error;
                     // reset view state
                     setViewState();
                 });
             }).catch((error) => {
-                CommonService.showToast('error', error.toString(), 20000);
+                error = error.toString();
+                if (error) {
+                    CommonService.showToast('error', error.toString(), 20000);
```
dfcowell: Why are we `toString`ing something we've already stringed two lines above?

app/src/templates/common/dialogs/choose-ledger-address.pug
```diff
@@ -0,0 +1,47 @@
...
+            .choose-title Select an Ethereum (ETH) address you would like to use:
+            .grid-container
+                md-radio-group(ng-model="selectedAddress")
+                    table
+                        thead
+                            tr
+                                td
+                                td Your Address
+                                td Balance
+                                td Token Balaces
```
altninja: typo: "Token Balances"

app/controllers/rpc-handler.js
```diff
@@ -977,6 +977,73 @@ module.exports = function (app) {
...
+                }
+            }).catch((error) => {
+                log.error(error);
+                app.win.webContents.send(RPC_METHOD, actionId, actionName, error, null);
+            });
+        } catch (e) {
+            log.error(e);
+            app.win.webContents.send(RPC_METHOD, actionId, actionName, e.message, null);
+        }
+    }
```
altninja: @dfcowell can we get the go-ahead to start a re-factor branch to improve how RPC is implemented?

app/src/templates/common/dialogs/choose-ledger-address.pug
altninja: typo: "Token Balances"

app/controllers/models/wallet.js
tmchedlishvili: Okay I will change it to 'enum'

app/src/app.js
```diff
@@ -212,6 +224,9 @@ angular.module('kyc-wallet').controller('GuestImportKeystoreController', GuestIm
 const GuestImportPrivateKeyController = requireAppModule('angular/controllers/guest/import/private-key-controller.js');
 angular.module('kyc-wallet').controller('GuestImportPrivateKeyController', GuestImportPrivateKeyController);
 
+const GuestImportLedgerController = requireAppModule('angular/controllers/guest/import/ledger.js');
+angular.module('kyc-wallet').controller('GuestImportLedgerController', GuestImportLedgerController);
+
```
altninja: this one too, there must be a better way to manage this otherwise the app will be completely unmaintainable after another 6 months of features being added

app/src/angular/services/web3.service.js
```diff
@@ -305,16 +253,13 @@ function Web3Service($rootScope, $window, $q, $timeout, $log, $http, $httpParamS
...
+        static waitForTicket(defer, method, args, contractAddress, contractMethod) {
+            RPCService.makeCall('waitForWeb3Ticket', { method, args, contractAddress, contractMethod }).then((response) => {
+                $log.info("method response", method, response);
+                defer.resolve(response)
+            }).catch((error) => {
+                $log.error("method response error", method, error);
+                defer.reject(error);
             });
         }
     };
```
altninja: question: why mix of Q and Bluebird for promises throughout the app?  can we refactor to native promises or async?

app/controllers/models/wallet.js
altninja: enum won't be supported in SQLite - will need to be string

app/src/app.js
tmchedlishvili: @altninja I don't understand this one, what should i do, every other controllers are implemented like this. You know, the issue is just about ledger integration

app/src/angular/configs/app.run.js
tmchedlishvili: This is a way to path some services into Token&Wallet classes, in order to access the services from the class. 
this is current, existing implementation if we need to change/refactor it, we should create another task/branch for this.
(this PR just stands for Ledger integration)

app/src/app.js
altninja: @tmchedlishvili its ok I'm not requesting changes here just commenting, it won't affect the PR

app/controllers/rpc-handler.js
dfcowell: Yeah, let's merge the open PRs and then we can get started on that, otherwise we'll be in merge hell. @altninja can you work through the PR backlog today please?

PR #188, opened 2018-05-09
SEL-27 dev done

tmchedlishvili:


app/controllers/common-service.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+const isOnline = require('is-online');
```
dfcowell: This file seems kinda silly when we can just import the `is-online` module wherever we need it.

tmchedlishvili:
@dfcowell common-service is removed.

PR #209, opened 2018-06-06
Fix/seed migration bugs

altninja:


package.json
```diff
@@ -31,7 +31,7 @@
   ],
   "main": "./src/main/main.js",
   "scripts": {
-    "dev": "gulp && electron-webpack dev",
+    "dev": "NODE_ENV=development gulp && electron-webpack dev",
```
rodrigopavezi: @altninja this should not be needed. Electron-webpack will set it when running the dev command. `electron-webpack dev` cheers

altninja: done

rodrigopavezi:
@altninja looks good 👍  cheers

PR #210, opened 2018-06-06
Fix/keystore unlock

altninja:


src/renderer/angular/controllers/guest/import/keystore-controller.js
```diff
@@ -98,6 +98,7 @@ function GuestImportKeystoreController($rootScope, $scope, $log, $q, $timeout, $
         });
 
         unlockExistingWalletPromise.then((data) => {
+            console.log('YES')
```
rodrigopavezi: @altninja can you remove this logging unless it is needed so it would be good to have more explanatory message. Cheers

altninja: oops missed one thanks!

rodrigopavezi:
@altninja looks good. cheers

PR #211, opened 2018-06-06
checkTxInfo: add guard into if statment

tmchedlishvili:


dfcowell:
@tmchedlishvili can you please fix the conflicts in this PR?

PR #215, opened 2018-06-12
Implement autoupdate

rodrigopavezi:


appveyor.yml
```diff
@@ -1,23 +1,44 @@
 image: Visual Studio 2017
 environment:
   OSENV: windows
-  OUTPUT: 'C:\projects\identity-wallet\sk-windows-' + $APPVEYOR_BUILD_VERSION + '.zip'
+  encryptedkey:
```
altninja: this is the same type of setup as the ones on Travis?  needs the keys from inside Appveyor to work right?

src/main/autoupdater.js
```diff
@@ -1,20 +1,12 @@
 'use strict'
-
-const { app, autoUpdater, dialog } = require('electron')
-const os = require('os')
-const updaterFeedURL = 'https://selfkey-releases.storage.googleapis.com/SelfKey-Desktop-Identity-Wallet/ea788838ccc9eb63676dd2da87772850/' + os.platform() + '/' + os.arch() + '/RELEASES.json?method=JSON&version=' + app.getVersion()
```
altninja: check with @dfcowell if we are still gonna use nucleus server

appveyor.yml
```diff
@@ -1,23 +1,44 @@
...
+  - ps: Start-Process 'Win64OpenSSL-1_1_0h.exe' -ArgumentList '/silent /verysilent /sp- /suppressmsgboxes' -Wait
+  - choco install gcloudsdk
+  - refreshenv
   - ps: Install-Product node 9 x64
   - npm i -g --production windows-build-tools
   - npm config set msvs_version 2015 --global
   - set PATH=C:\Ruby22\bin;%PATH%
   - gem install sass compass
   - npm i -g yarn
-  - npm i --no-prune
```
altninja: this was breaking builds without it - if yarn solves then cool but for me it caused a lot of dependency issues

appveyor.yml
rodrigopavezi: @altninja this is different. the values on the yml are encrypted and appveyor will decrypt it when running a push build. That is why the value is declared here. I am using the same travis keys  though. I deploy their values to GCS so I could also setup here and on cricle ci. cheers

appveyor.yml
rodrigopavezi: @altninja yeah, with yarn there is not need for it. cheers

src/main/autoupdater.js
rodrigopavezi: @altninja good point. I have setup auto update with github release. I guess we can point a custom domain to it from the selfkey website. Something like this https://help.github.com/articles/linking-to-releases/ 

@dfcowell thoughts?

Cheers

dfcowell: Let's keep using Github releases. Less to maintain. :+1:

PR #216, opened 2018-06-12
Fix SEL-49 by doing the right formatting and unformatting of the values

rodrigopavezi:


src/renderer/angular/directives/commons/sk-circle-pie-chart.directive.js
```diff
@@ -39,14 +39,14 @@ function SkCirclePieChartDirective($timeout, CommonService) {
 
                 scope.totalValueFormated = new BigNumber(0);
                 items.forEach((item) => {
-                    item.valueUSD = Number(CommonService.numbersAfterComma(item.valueUSD, 2));
-                    scope.totalValueFormated = scope.totalValueFormated.plus(new BigNumber(item.valueUSD));
+                    item.unformattedValueUSD = CommonService.formattedLocaleNumberToNumber(item.valueUSD);
```
dfcowell: If we have to do this we're formatting too early.

There should only be a single conversion, from unformatted to formatted, as late as possible, just before the value is rendered to the screen.

src/renderer/angular/directives/commons/sk-circle-pie-chart.directive.js
```diff
@@ -64,14 +64,15 @@ function SkCirclePieChartDirective($timeout, CommonService) {
...
                         color: colorForOthers
                     };
 
                     otherItems.forEach(otherItem => {
-                        otherAggregated.valueUSD = new BigNumber(otherAggregated.valueUSD).plus(new BigNumber(otherItem.valueUSD));
+                        otherAggregated.unformattedValueUSD = new BigNumber(otherAggregated.unformattedValueUSD).plus(new BigNumber(otherItem.unformattedValueUSD));
                     });
 
-                    otherAggregated.valueUSD = +CommonService.numbersAfterComma(otherAggregated.valueUSD, 2);
+                    otherAggregated.valueUSD = +Intl.NumberFormat('en-US').format(Number(otherAggregated.unformattedValueUSD).toFixed(2));
```
dfcowell: Passing USD attributes around is going to make implementing localized currencies hell.

We should pass around a standard `fiatValue` attribute which contains a number which may represent any fiat currency.

rodrigopavezi:
@dfcowell I have fix it from the top but I reckon I can work on a react component for sure. If there isn't any other priority for the release date. I can work on the new component. Thoughts?

dfcowell:
@rodrigopavezi looks good. I'll merge this, but let's create a new ticket and branch for implementing it as a React component and you can start working on it when you have time.

PR #218, opened 2018-06-13
Fix Issue with screen for adding Attributes and Documents plus an iss…

rodrigopavezi:
…ue with Prices


src/main/controllers/cmc-service.js
```diff
@@ -56,8 +56,10 @@ module.exports = function (app) {
                         return callback(err);
                     });
                 }, function (err) {
-                    if (err) return log.error(err);
-                    log.info("TOKEN PRICE UPDATED");
+                    if (err) {
```
dfcowell: This looks like debugging/development code. Do we need it in the production app?

src/main/controllers/cmc-service.js
```diff
@@ -56,8 +56,10 @@ module.exports = function (app) {
...
+                        log.error('TOKEN PRICE UPDATE', err)
+                    }
+                    log.error('TOKEN PRICE UPDATED');
                     eventEmitter.emit('UPDATE');
                     setTimeout(loadCmcData, config.cmcUpdatePeriod);
```
dfcowell: Would it be worth refactoring this system to use `setInterval` instead?

PR #219, opened 2018-06-14
Refactor knex and seed implementation. Add error handling for migrati…

dfcowell:
…on errors.


package.json
```diff
@@ -37,7 +37,8 @@
     "dist": "yarn compile && electron-builder",
     "dist:dir": "yarn dist --dir -c.compression=store -c.mac.identity=null",
     "test": "node test/test.js",
-    "publish-build": "yarn compile && electron-builder -p always"
+    "publish-build": "yarn compile && electron-builder -p always",
+    "rebuild": "npm rebuild && ./node_modules/.bin/electron-rebuild -f -v 2.0.2"
```
rodrigopavezi: @dfcowell I think you added this because you were not able to run yarn dev. That is because before running yarn dev you have to run yarn install-app-deps. The install app deps was a yarn postinstall script that would run automatically. I had to disable it for now as it was breaking windows build with the current version of electron builder. This has been fixed on the newer electron builder version but was not released yet and has another bug. So once it is stable I will update so we can have the postinstall back. For now we need to remember to run it after running yarn install. Cheers

rodrigopavezi:
@dfcowell I just made a couple of changes plus fixed the bug. Let me know if you don't agree with them. Cheers

dfcowell:
@rodrigopavezi only thing is that I added a menu that appears in debug and development modes that lets us open devtools, so I don't know if the automatic popping open is needed any more. Not a big deal though.

PR #226, opened 2018-06-19
Fix/sel 21

rodrigopavezi:


src/main/menu.js
```diff
@@ -12,7 +12,7 @@ const getMenuTemplate = (mainWindow) => {
                 {
                     label: "About", role: 'about',
                     click() {
-                        const win = new electron.BrowserWindow({
+                        let win = new electron.BrowserWindow({
```
dfcowell: There's nothing that requires this to be a `let` - we should prefer `const` declarations where possible.

rodrigopavezi: @dfcowell I know but `win = null` when closing the window requires it to be a let.  Cheers

dfcowell: D'oh! Didn't see that.

PR #230, opened 2018-06-19
Change token hide to only hide from the chart on the dashboard

rodrigopavezi:


src/renderer/angular/controllers/member/dashboard/main-controller.js
```diff
@@ -67,21 +67,22 @@ function MemberDashboardMainController($rootScope, $scope, $interval, $log, $q,
         Object.keys(wallet.tokens).forEach((tokeyKey) => {
             let pieChartItem = {};
             let token = wallet.tokens[tokeyKey];
+            if (!token.isHidden()) {
```
dfcowell: These negative conditionals can be difficult to read and are prone to errors.

In a case like this, we should usually write this as a guard clause with an early return to prevent processing.

```
if (token.isHidden()) {
  return;
}
```

rodrigopavezi: @dfcowell I don't think the return will work inside a forEach. I just need to skip the token that has hidden equal true from being added to the list. I can't see another way of doing it. Cheers

dfcowell: A callback in a `forEach` is like any other function. It will work.

It won't work in a `map`, since it would leave a `null` value in the resulting array.

rodrigopavezi: @dfcowell nvm, I think it will apply the continue.

rodrigopavezi:
@dfcowell good call. I have changed it. cheers

PR #233, opened 2018-06-21
Fix sending eth, key and custom token

rodrigopavezi:
This PR contains fixes for sending eth, key and custom tokens.

It specifically fixing the issue where after a transaction is sent the get transaction count does not return the count with the recent transaction sent. So I added a retry until it gets the right count to then create a new signed transaction.

Cheers

### DIFF
--- a/src/renderer/angular/classes/wallet.js
+++ b/src/renderer/angular/classes/wallet.js
@@ -4,7 +4,7 @@ const EthUnits = require('./eth-units');
 const EthUtils = require('./eth-utils');
 const Token = require('./token');
 
-let $rootScope, $q, $interval, Web3Service, CommonService, ElectronService, SqlLiteService, EtherScanService, SignService;
+let $rootScope, $q, $interval, Web3Service, CommonService, ElectronService, SqlLiteService, EtherScanService, SignService, $log;
 
 let readyToShowNotification = false;
 
@@ -21,6 +21,7 @@ class Wallet {
     static set SqlLiteService(value) { SqlLiteService = value; }
     static set EtherScanService(value) { EtherScanService = value; }
     static set SignService(value) { SignService = value; }
+    static set $log(value) { $log = value; }
     
 
     constructor(id, privateKey, publicKey, keystoreFilePath, profile) {
@@ -53,6 +54,8 @@ class Wallet {
         this.initialBalancePromise = this.loadBalance();
 
         this.loadTokens();
+
+        this.previousTransactionCount = -1;
     }
 
     getPrivateKey() {
@@ -250,6 +253,17 @@ class Wallet {
         let promise = Web3Service.getTransactionCount(this.getPublicKeyHex());
         promise.then((nonce) => {
             //wallet.nonceHex
+
+            if (nonce <= this.previousTransactionCount) {
+                this.generateRawTransaction(toAddressHex, valueWei, gasPriceWei, gasLimitWei, contractDataHex, chainID).then(res => {
+                    defer.resolve(res)
+                }).catch(err => {
+                    defer.reject(err);
+                });
+                return defer.promise;
+            } 
+
+            this.previousTransactionCount = nonce;
 
             let rawTx = {
                 nonce: EthUtils.sanitizeHex(EthUtils.decimalToHex(nonce)),
--- a/src/renderer/angular/classes/wallet.js
+++ b/src/renderer/angular/classes/wallet.js
@@ -23,7 +23,6 @@ class Wallet {
     static set SignService(value) { SignService = value; }
     static set $log(value) { $log = value; }
     
-
     constructor(id, privateKey, publicKey, keystoreFilePath, profile) {
         this.id = id;
         this.keystoreFilePath = keystoreFilePath;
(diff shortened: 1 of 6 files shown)


src/renderer/angular/classes/wallet.js
```diff
@@ -251,6 +254,17 @@ class Wallet {
         promise.then((nonce) => {
             //wallet.nonceHex
 
+            if (nonce <= this.previousTransactionCount) {
+                this.generateRawTransaction(toAddressHex, valueWei, gasPriceWei, gasLimitWei, contractDataHex, chainID).then(res => {
```
dfcowell: Is there some reason we're not just using a real promise here @rodrigopavezi?

Seems like we could just `return` on this line.

src/renderer/angular/controllers/commons/dialogs/send-token-controller.js
```diff
@@ -278,6 +283,9 @@ function SendTokenDialogController($rootScope, $scope, $log, $q, $mdDialog, $int
             error = error.toString();
             if (error.indexOf('Insufficient funds') == -1) {
                 CommonService.showToast('error', error, 20000);
+            } else if(error.indexOf('Endpoint request timed out') !== -1){
+                sendEther();
```
dfcowell: The `sendToken` and `sendEther` functions look exactly the same to me.

Could we refactor to just use one function for sending both?

src/renderer/angular/classes/wallet.js
```diff
@@ -4,7 +4,7 @@ const EthUnits = require('./eth-units');
 const EthUtils = require('./eth-utils');
```
nectos: Only one require is needed for imports.

src/renderer/angular/controllers/commons/dialogs/send-token-controller.js
```diff
@@ -92,6 +92,11 @@ function SendTokenDialogController($rootScope, $scope, $log, $q, $mdDialog, $int
                 }
             }
 
+            if(error.indexOf('Endpoint request timed out') !== -1){
```
nectos: I'd highly suggest not to have hardcoded string checks in code, move to ``constansts`` instead.
When needed, you could change constant instead of searching codebase for ``strings``

src/renderer/angular/controllers/commons/dialogs/send-token-controller.js
```diff
@@ -301,6 +309,9 @@ function SendTokenDialogController($rootScope, $scope, $log, $q, $mdDialog, $int
             error = error.toString();
             if (error.indexOf('Insufficient funds') == -1) {
                 CommonService.showToast('error', error, 20000);
+            }else if(error.indexOf('Endpoint request timed out') !== -1){
```
nectos: need additional space before ``else``

src/renderer/angular/configs/app.run.js
```diff
@@ -36,6 +36,8 @@ function AppRun($rootScope, $log, $window, $timeout, $interval, $q, $state, $tra
     Wallet.SqlLiteService = SqlLiteService;
     Wallet.CommonService = CommonService;
     Wallet.EtherScanService = EtherScanService;
+    Wallet.$log = $log;
+
```
nectos: Delete extra line.

src/renderer/angular/controllers/commons/dialogs/send-token-controller.js
nectos: Just noticed that there are 8 findings of ``error_while_creating`` inside ``src/main/controllers/sql-lite-service.js``.

src/renderer/angular/controllers/commons/dialogs/send-token-controller.js
```diff
@@ -301,6 +309,9 @@ function SendTokenDialogController($rootScope, $scope, $log, $q, $mdDialog, $int
             error = error.toString();
```
nectos: Just wondering, what structure ``error`` object has?

nectos:
In general ``No description provided.`` on Pr's is not good. I bet you have Jira or Trello or whatever, but you don't tag issues created there, nor you tell what is this PR fixing. If you guys are looking at this project seriously, please do not trash the commit history with no explanation.

src/renderer/angular/classes/wallet.js
rodrigopavezi: @dfcowell I figured now a better way. I will need to pass the defer to the recursive function call. Then I can just return it here. Cheers

src/renderer/angular/controllers/commons/dialogs/send-token-controller.js
rodrigopavezi: Yeap, will refactor. Cheers

src/renderer/angular/controllers/commons/dialogs/send-token-controller.js
```diff
@@ -5,6 +5,8 @@ const EthUtils = require('../../../classes/eth-utils');
 
 function SendTokenDialogController($rootScope, $scope, $log, $q, $mdDialog, $interval, $window, CONFIG, $state, $stateParams, Web3Service, CommonService, SqlLiteService) {
     'ngInject'
+
+    const TIMEOUT_ERROR = "Endpoint request timed out";
```
nectos: this should not be in ``function`` scope.

src/renderer/angular/controllers/commons/dialogs/send-token-controller.js
```diff
@@ -267,7 +265,7 @@ function SendTokenDialogController($rootScope, $scope, $log, $q, $mdDialog, $int
         return txGenPromise;
     }
 
-    function sendEther() {
+    function send() {
```
nectos: Please consider not to name functions in so basic way. ``sendSmartContract`` instead?

rodrigopavezi:
@dfcowell I have made the refactoring. Please review. Cheers

PR #238, opened 2018-06-25
WIP: refactor(data-layer): update data sources & remove excess code

dfcowell:
This commit updates the wallet to use coincap.io for token/coin prices. It also removes excess or
dead code in the data layer (sqlite/airtable). cmc-service has been renamed to price-service and
initializing models/services has been simplified where possible.


src/main/controllers/price-service.js
```diff
@@ -0,0 +1,72 @@
...
+
+const EventEmitter = require('events');
+const fetch = require('node-fetch');
+
+const TokenPrice = require('./models/token-price');
+
+const emitter = new EventEmitter();
+
+const loadPriceData = async () => {
+	const response = await fetch('https://coincap.io/front');
```
rodrigopavezi: @dfcowell I was thinking to use the streaming api https://github.com/CoinCapDev/CoinCap.io#socket-api. so we won't need a refresh and we could even not need to store it on the DB as the wallet cannot run offline. Thoughts? cheers

dfcowell: I like the idea of the streaming API however we should keep a local cached copy in order to speed up wallet start time (we don't want to have to wait to download the token list or receive it through the stream before we can render balance info) and provide gracefully degraded functionality in the event that the CoinCap API goes down.

rodrigopavezi: Agreed about caching locally though. Do you know if coins update all together? If yes, there is no saving burden when editing on the DB. Otherwise if they change separated it should increase performance as we don't need to edit 13 hundreds records at once.

src/main/controllers/models/token-price.js
```diff
@@ -1,63 +1,26 @@
...
+
+	findBySymbol: symbol =>
+		knex(TABLE_NAME)
+			.select()
+			.where({ symbol })
+			.then(rows => (rows && rows.length ? rows[0] : null)),
+
+	add: tokenPrice => knex(TABLE_NAME).insert(tokenPrice),
+
+	edit: tokenPrice =>
```
rodrigopavezi: @dfcowell Curious, is there a reason why the edit is not using transaction? Cheers

dfcowell: I was in a hurry and forgot to implement transactions. 😄 This is why we do code reviews!

src/main/controllers/models/token-price.js
```diff
@@ -1,63 +1,26 @@
...
+		knex(TABLE_NAME)
+			.where('id', tokenPrice.id)
+			.update(tokenPrice),
+
+	bulkEdit: tokenPrices => Promise.all(tokenPrices.map(controller.edit)),
+
+	bulkAdd: tokenPrices => knex.batchInsert(TABLE_NAME, tokenPrices)
```
sk91: @dfcowell Batch insert fails with "too many SQL variables"
When bringing back the bulk add from sqlLiteService it works
I think we should refactor out from sqlLiteService sql util functions and use them in models

rodrigopavezi:
@dfcowell can we merge this one? cheers

dfcowell:
Closing since #247 includes this work.

PR #242, opened 2018-06-26
feat: add Crash Reporter with sentry

rodrigopavezi:
Implement the sentry electron crash report. If user opt-in it will send exceptions and crash reports to sentry so we can fix bugs improving the wallet


src/main/controllers/rpc-handler.js
```diff
@@ -676,6 +677,8 @@ module.exports = function(app) {
 		}
 
 		function getDocumentRequirements(kycprocess) {
+			process.crash();
```
rodrigopavezi: this is for testing purpose. It will be removed. cheers

src/main/controllers/crash-report-service.js
```diff
@@ -0,0 +1,30 @@
...
+const log = require('electron-log');
+// DSN will be inject during build time
+const sentry = require('../assets/data/sentry');
+let hasStarted = false;
+
+module.exports = {
+	startCrashReport: async () => {
+		if (process.env.NODE_ENV !== 'development' && process.env.MODE !== 'test') {
+			const result = await guideSettings.findAll();
+			if (result[0].crashReportAgreement && !hasStarted) {
```
dfcowell: This will work most of the time, but Sentry also offers the `shouldSend` callback which is more reliable and cleaner. I know you've added the "startCrashReport" call as a side effect when updating the settings, but that violates the SRP (updating the database shouldn't start a service) and tightly couples the settings handler and our Sentry implementation.

https://docs.sentry.io/clients/electron/config/#options

PR #244, opened 2018-06-26
fix reload tx-history progress bar issue

tmchedlishvili:


src/renderer/angular/directives/commons/sk-tx-history.directive.js
```diff
@@ -1,117 +1,128 @@
...
+			$rootScope.$on('balance:change', event => {
+				loadData();
+			});
+
+			$rootScope.$on('tx-history:change', event => {
+				loadData();
+			});
+
+			$rootScope.$on('tx-history:sync', event => {
+				scope.isSyncing = true;
```
tmchedlishvili: @dfcowell only this line is added.

dfcowell: Where are all of the other changes coming from?

tmchedlishvili: In this file, it comes from `tab space issue`

PR #245, opened 2018-06-26
fix: add handle for transaction receipt error

rodrigopavezi:
Change to ignore the error as it looks like to be intermittent and won't impact the sending of the transaction.


dfcowell:
@rodrigopavezi are we sure that there aren't cases where this error is legitimate?

There appear to be other workarounds on that Github issue you linked.

If 80% of the time it's a false error, but 20% of the time it's real our wallet will look buggy that 20% of the time.

Can we update the `Web3Service.sendRawTransaction` implementation to only return that error if we don't have a confirmation/hash from web3?

rodrigopavezi:
closing this PR as @tmchedlishvili said it should be fixed with his changes to the transaction.

PR #247, opened 2018-06-27
refactor(sql util): refactor out common sql utils

sk91:
@dfcowell 
This commit moves common sql function into the knex service so we can use it anywhere simply by requiring them.

Also it fixes bulk Insert of token price.


dfcowell:
Can you take a look at this PR too @sk91? Maybe you can incorporate some of the work I've already done into what you're working on. https://github.com/SelfKeyFoundation/Identity-Wallet/pull/238

sk91:
Hi, my branch is a fork of this branch, did it on a separate branch since
it's my first commit.

ср, 27 июн. 2018 г. в 13:28, Dan Cowell <notifications@github.com>:

src/main/migrations/20180508093508_init.js
```diff
@@ -8,14 +8,20 @@ exports.up = function(knex, Promise) {
 				.references('wallets.id');
 			table.string('title');
 			table.string('content');
-			table.integer('createdAt').notNullable();
+			table
```
dfcowell: We can't change this file now that people have run the wallet.

Create a new migration to make these schema changes using the `knex` CLI.

src/main/services/knex.js
```diff
@@ -39,7 +39,132 @@ const init = async () => {
 	}
 };
 
+const sqlUtil = {
```
dfcowell: The vast majority of these functions are a bad idea.

There's no compelling reason to wrap the `knex` API in this layer of indirection. It makes debugging harder because we're adding our own custom API on top (is the bug in our helpers, in `knex` or in how we implement the `knex` API?) and it's a very leaky abstraction, since the calling code needs to know about how `knex` will handle the transaction object passed in.

We should get rid of as many of these functions as possible and prefer using the `knex` builder where possible - or implement something like Bookshelf.

sk91: Will move everything to objection.js

PR #255, opened 2018-07-02
Fix/sel 13

tmchedlishvili:


src/main/controllers/ledger-service.js
```diff
@@ -1,151 +1,100 @@
...
-	function obtainPathComponentsFromDerivationPath(derivationPath) {
-		// check if derivation path follows 44'/60'/x'/n pattern
-		const regExp = /^(44'\/6[0|1]'\/\d+'?\/)(\d+)$/;
-		const matchResult = regExp.exec(derivationPath);
-		if (matchResult === null) {
-			throw new Error(
-				"To get multiple accounts your derivation path must follow pattern 44'/60|61'/x'/n "
-			);
+		if (err instanceof TimeoutError) {
+			return reject('CUSTOM_TIMEOUT');
```
dfcowell: If `reject` is a promise rejection we should only ever pass `Error` objects into it.

src/main/controllers/ledger-service.js
```diff
@@ -1,151 +1,100 @@
...
-		const pathComponents = obtainPathComponentsFromDerivationPath(derivationPath);
-
-		let addresses = {};
-		for (let i = indexOffset; i < indexOffset + accountsNo; i += 1) {
-			let path = pathComponents.basePath + (pathComponents.index + i).toString();
-			let address = await eth.getAddress_async(path);
-			addresses[path] = address.address;
+		let message = (err.message || '').toLowerCase();
+		if (message.indexOf('denied by the user') != -1) {
+			return reject('DENIED_BY_THE_USER');
```
dfcowell: Just pass in the `err` object here. Why are we creating our own less-useful error codes?

src/main/controllers/ledger-service.js
```diff
@@ -1,151 +1,100 @@
...
 		}
 
-		return addresses;
+		reject(errorsMap[message] || 'UNKNOWN_ERROR');
```
dfcowell: Pass an `Error` object.

src/main/controllers/ledger-service.js
```diff
@@ -1,151 +1,100 @@
...
+		let signPromise = electron.app.web3Service.waitForTicket({
+			method: 'signTransaction',
+			args: [args.dataToSign],
+			contractAddress: null,
+			contractMethod: null,
+			onceListenerName: null,
+			customWeb3: createWeb3()
+		});
 
+		return new Promise((resolve, reject) => {
```
dfcowell: There's no need for this. There's clearly a promise immediately below this. Just return that one.

src/main/controllers/ledger-service.js
```diff
@@ -1,151 +1,100 @@
...
-					}
-					reject(err.toString());
+					onError(err, reject);
 				});
 		});
-	};
+	}
+
+	async function _getAccounts(args) {
+		return new Promise((resolve, reject) => {
```
dfcowell: This is pointless. Just `return` the call to `waitForTicket`. It returns a promise.

src/main/controllers/ledger-service.js
```diff
@@ -1,151 +1,100 @@
...
+			electron.app.web3Service
+				.waitForTicket({
+					method: 'getAccounts',
+					args: [],
+					contractAddress: null,
+					contractMethod: null,
+					onceListenerName: null,
+					customWeb3: createWeb3(args.start, args.quantity)
+				})
+				.then(result => {
```
dfcowell: This does nothing. Delete it.

src/main/controllers/ledger-service.js
```diff
@@ -1,151 +1,100 @@
...
+					resolve(result);
+				})
+				.catch(err => {
```
dfcowell: This is also useless.

src/renderer/angular/controllers/commons/dialogs/send-token-controller.js
```diff
@@ -128,16 +111,16 @@ function SendTokenDialogController(
 				}
 
 				if (isLedgerWallet) {
```
dfcowell: `switch`/`case`

src/renderer/angular/services/ledger.service.js
```diff
@@ -73,12 +71,11 @@ function LedgerService($log, RPCService, CommonService, Web3Service) {
 			return RPCService.makeCall('createLedgerWalletByAdress', { address });
 		}
 
-		signTransaction(dataToSign, address, derivationPath) {
+		signTransaction(dataToSign, address) {
 			return new Promise((resolve, reject) => {
```
dfcowell: Get rid of this useless promise wrapper.

PR #257, opened 2018-07-03
feat(tests): add basic unit test

sk91:
Add distinction between e2e and unit tests and setup basic unit tests


dfcowell:
@rodrigopavezi @sk91 I'm fine with moving to Jest for unit tests as long as it can also run our e2e tests without requiring a massive refactor.

If we need to rewrite all of our existing tests it's not really worth it.

rodrigopavezi:
@dfcowell agreed. I was thinking to use jest only for unit test but maybe it is not even worth for that. cheers

PR #262, opened 2018-07-05
fix(terms): fix scroll to bottom not triggering in some cases

sk91:
Not tested it on a customer yet

### DIFF
--- a/src/renderer/angular/controllers/commons/dialogs/terms-controller.js
+++ b/src/renderer/angular/controllers/commons/dialogs/terms-controller.js
@@ -1,6 +1,6 @@
 'use strict';
 
-function TermsDialogController($rootScope, $scope, $log, $q, $mdDialog, SqlLiteService) {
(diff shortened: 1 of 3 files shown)


sk91:
Tested with customers, the fix seems to work

src/renderer/angular/controllers/commons/dialogs/terms-controller.js
```diff
@@ -41,8 +41,9 @@ function TermsDialogController($rootScope, $scope, $log, $q, $mdDialog, SqlLiteS
 
 	$scope.scrollToEndContainer = direction => {
 		if (direction === 'bottom') {
-			$scope.scrolledBottom = true;
-			$scope.$apply();
+			$timeout(() => {
```
dfcowell: What does this timeout do, @sk91?

sk91: @dfcowell it does scope.$apply in a safe way ($scope.apply will fail if digest cycle is in progress)

Popular prectice in angular

PR #263, opened 2018-07-05
feat: implement integration of react with redux

rodrigopavezi:


src/common/locale/types.js
```diff
@@ -0,0 +1,3 @@
+const LOCALE_UPDATE = 'locale_update';
```
dfcowell: Note the "namespacing" of action types in the example here: https://github.com/alexnm/re-ducks

This action type should be "app/locale/UPDATE" or "app/locale/UPDATE_LOCALE"

src/main/preload.js
```diff
@@ -15,6 +15,15 @@ require('angular-sanitize');
 require('angular-local-storage');
 require('angular-qrcode');
 require('angular-zxcvbn');
+window.react = require('react');
```
dfcowell: Why are we attaching these to `window` instead of `require`ing them where necessary?

src/main/preload.js
```diff
@@ -42,10 +51,23 @@ window.requireAppModule = function(moduleName, isNear) {
 };
 
 window.requireNodeModule = function(moduleName) {
```
dfcowell: What's the purpose of this?

src/renderer/react/priceBox/components/price-box.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { CryptoPriceBox } from 'selfkey-ui';
+
+class PriceBox extends React.Component {
```
dfcowell: What is this component for?

src/renderer/react/priceBox/index.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+import { getInitialStateRenderer } from 'electron-redux';
+import configureStore from 'common/configure-store';
+import PriceBox from './containers/price-box';
+
+const initialState = getInitialStateRenderer();
+const store = configureStore(initialState, 'renderer');
```
dfcowell: Why are we configuring the store in a component file? We should do it when we initialise the store for the first time.

src/renderer/react/priceBox/containers/price-box.js
```diff
@@ -0,0 +1,18 @@
...
+};
+
+const mapDispatchToProps = dispatch => {
+	return bindActionCreators(localeActions, dispatch);
+};
+
+export default connect(
+	mapStateToProps,
+	mapDispatchToProps
+)(PriceBox);
```
dfcowell: We could just use the `CryptoPriceBox` from `selfkey-ui` here.

src/main/preload.js
rodrigopavezi: @dfcowell that is because of the `nodeIntegration` false. It won't require the dependency in run time. I mean on the renderer process. So I need to require it on main process through the preload. That sucks I know but that I think it the way to go if we want more security.

src/main/preload.js
rodrigopavezi: @dfcowell as explained above we need to load the dependencies here so they will be accessible to renderer process.

src/renderer/react/priceBox/components/price-box.js
rodrigopavezi: @dfcowell good point, we can remove it and move the CryptoPrixeBox directly to the container. cheers

src/main/preload.js
dfcowell: Why does a simple `require` need `nodeIntegration` if we're bundling it with Webpack? This is what webpack is designed for.

rodrigopavezi: @dfcowell yeah, it works differently with electron. With  nodeIntegration false it want allow any code from the renderer process to make a require to node modules. Only to modules on with code inside the renderer process. Otherwise anyone could require through the dev tools console or inject a require code somewhere. So it protects from that. Usually electron apps don't need to set nodeIntegration false so it is really hard to find material on the internet on how to handle node modules on renderer process when it is false. I tried many different ways to get the modules loaded but that was the only way that it worked. Cheers

dfcowell: There is no way that the only way to load modules in the renderer in Electron is to bind them all to the window object.

It's a web browser. If we run webpack to build an app bundle there's no reason it wouldn't work. I'll fix this one myself.

src/main/preload.js
rodrigopavezi: @dfcowell It might not need to bound to window but previously it was so I just kept like that. But I quite sure there is no way to just not load the required modules on the preload and handle that over to the renderer process when need with the switch I coded there. Cheers

rodrigopavezi: @dfcowell btw, require on rendering process works fine when on `yarn dev` mode but not when building the app with electron builder, `yarn dist` then `open dist/mac/SelfKey\ Identity\ Wallet.app/`. Just so you make sure to build it when testing any change. cheers

src/main/preload.js
dfcowell: @rodrigopavezi I've done a lot more research. We have two options:

1. *Set `nodeIntegration` to `true`:* There's no reason we need it set to `false` for the main window of the application. All that runs in that window is local, trusted code. I prefer this option.
2. *Extend `electron-webpack` as a library and override its `target` for the renderer build, making it treat the renderer like a standard web SPA - which would mean bundling its own `require` and all of the various modules. This adds a lot of maintenance overhead, but we would be able to very strictly control what the renderer has access to.

I think we should only go with option 2 if there is a compelling reason, which I personally can't find.

src/main/preload.js
rodrigopavezi: @dfcowell it is not only us that have this issue https://stackoverflow.com/questions/49705639/configuring-electron-webpack-renderer-to-work-with-nodeintegration-false

Cheers

src/main/preload.js
rodrigopavezi: @dfcowell this guide has the reason why we have chosen to set nodeIntegration to false.

dfcowell: That specifically says "for untrusted origins."

It's specifically talking about loading http://some-remote-untrusted-webpage.com/whatever in an electron context.

We don't do that.

rodrigopavezi: @dfcowell good point, so we are just over thinking here and being too protective in a way that hurt us :-) funny. Okay I will set it to true and remove all this preload require stuff. Do you agree? Cheers

dfcowell:
@rodrigopavezi do you want to continue on this branch for the rest of the dashboard improvements and fixes? We still need to move the crypto exchange prices into Redux and update the rest of the dashboard UI components that display crypto value in dollars. We can either do that as part of this PR, or merge what you've got now and continue with a new branch.

rodrigopavezi:
@dfcowell good question, I would rather merge this one and create small ones for each currency UI component we need to build. After I get all the UI done I can create the redux middleware for getting the prices from the data layer.

dfcowell:
@rodrigopavezi noted. Please resolve the conflicts and feel free to merge. We really need to come up with a solution for these e2e test failures eventually, too.

rodrigopavezi:
@dfcowell yeah, I have two things in mind to improve on that.

1. Find a more verbose way of running it so we can see what is going on on the ci env. Perhaps have even screenshots when it fails.
2. Move this Mac build to Circle ci. That should increase our velocity as it will build faster and we can split the test in a separated job which can be re run without having to re run the whole build. Downside is that we would need to pay for mac in circle ci.  

Thoughts?

Cheers

dfcowell:
@rodrigopavezi I'm all for simplifying our tooling, so if we can standardise on CircleCI that would be great. We still need Appveyor for Windows builds though, right?

rodrigopavezi:
@dfcowell yeap, unfortunately we still need Appveyor but I hope not for long as CircleCI has windows support on their backlog.  I will take a look on Circle Ci for mac if that does not take too long I will work on switching to it. Cheers

Issue #268, opened 2018-07-09
Data Persistence Layer: Decouple & Improve Test Coverage 


dfcowell:
- [x] add unit tests #257
- [x] migrate to objection.js #247 
- [x] cover all models in tests #278
- [x] integrate jest tests #295 
- [x] refactor id attributes #361 
- [x] refactor services #316
- [x] require -> import #316
- [x] add test coverage report #371

sk91:
#257 
#247 
#277

PR #278, opened 2018-07-10
feat(model-tests): cover all models with tests

sk91:
#268 feat to cover all models in tests

- [x] id-attributes
- [x] tx-history
- [x] wallet-settings
- [x] token price
- [x] token
- [x] wallet token
- [x] wallet
- [x] finish import for id attribute type


.editorconfig
```diff
@@ -5,6 +5,6 @@ root = true
 end_of_line = lf
 charset = utf-8
 indent_size = 4
-indent_style = space
+indent_style = tab
```
rodrigopavezi: @sk91 curious, why use tab instead of spaces?

sk91:
@rodrigopavezi That’s how our prettier (in .prettierrc) is configured, just
resolved a conflict

src/main/models/wallet-setting.js
```diff
@@ -18,7 +18,7 @@ class WalletSetting extends BaseModel {
 			properties: {
 				id: { type: 'integer' },
 				walletId: { type: 'integer' },
-				showDesktopNotifications: { type: 'integer' },
+				sowDesktopNotifications: { type: 'integer' },
```
rodrigopavezi: @sk91 I think there is a typo here.

sk91: @rodrigopavezi should we care about getting this fixed?

This is a regression. Happened long time ago in initial db schema. Bubbled up to new models.
Sqlite does not support column rename support. To rename it now we need to drop the whole table and create from scratch:

Migration will look like this:

```
exports.up = async (knex, Promise) => {
	// rename sowDestopNotifications to showDesktopNotifications
	await knex.schema.renameTable('wallet_settings', 'wallet_settings_bkp');
	await knex.schema.createTable('wallet_settings', table => {
		table.increments('id');
		table
			.integer('walletId')
			.notNullable()
			.references('wallets.id');
		table
			.integer('showDesktopNotifications')
			.notNullable()
			.defaultTo(0);
		table.string('airDropCode');
		table
			.integer('txHistoryLastSyncedBlock')
			.notNull()
			.defaultTo(0);
		table.integer('createdAt').notNullable();
		table.integer('updatedAt');
	});
	let data = await knex('wallet_settings_old').select(
		'id',
		'walletId',
		'sowDesktopNotifications as showDesktopNotifications',
		'airDropCode',
		'txHistoryLastSyncedBlock',
		'createdAt',
		'updatedAt'
	);
	await knex('wallet_settings').insert(data);
};

exports.down = async (knex, Promise) => {};
```

dfcowell:
Couldn’t we create a column, move the data, and drop the incorrect column?

sk91:
@dfcowell 
> SQLite supports a limited subset of ALTER TABLE. The ALTER TABLE command in SQLite allows the user to rename a table or to add a new column to an existing table. It is not possible to rename a column, remove a column, or add or remove constraints from a table.

rodrigopavezi:
@dfcowell @sk91 that is a tough call but I think we should not fix it as it will make the wallet slower for first time users. I mean, every time we add a migration we are slowing down the startup of the wallet.  I think we should save in migrations for when we really need to migrate something more important. Thoughts?

sk91:
@dfcowell @rodrigopavezi 

On another thought, I believe I can fix this on a model level, but:
I don't see this option used anywhere besides, maybe I should Just hide it

sk91:
@dfcowell @rodrigopavezi 
Fixed on model level (tranlating showDesktopNotifications to sowDesktopNotificaitons and back)

PR #284, opened 2018-07-11
fix: ledger connection

tmchedlishvili:


tmchedlishvili:
@rodrigopavezi 
My first implementation was `const async = require('async')`. 
it's changed/reverted again.

Issue #269, opened 2018-07-09
Implement: Local Error Log with Logfile Export


dfcowell:
Local log should support:

- Multiple log levels
- The ability to specify a log level via environment variable or command line argument
- The ability to export to a text file via a menu item in the application menu bar
- Mark start point for issue reproduction via menu item

Progress:

- [x] Select and integrate logging library (is electron-log good enough?) #281 
- [x] Define log format #281 
- [x] Configure log levels via env vars #281 
- [x] use the new logger in all codebase #291 
- [x] Add option to export log file #301 
- [x] Mark issue for reproduction #293 
- [x] Add option to change to debug log levels on demand #301 
- [x] integrate with sentry #301

sk91:
> Mark start point for issue reproduction via menu

@dfcowel what do you mean by that?

sk91:
> Mark start point for issue reproduction via menu

@dfcowell what do you mean by that?

dfcowell:
Sorry @sk91 I missed that.

Log files (especially very detailed ones) can get very long.

If we are communicating with a user and they send through a log file it can be time consuming trying to find the relevant part of the file.

Allowing the user to "mark" the file means giving them a menu option that when selected logs a distinctive message to the file like "SUPPORT: ISSUE REPRODUCTION STARTS HERE"

Imagine the case where a user says that they are having trouble adding a token to their wallet. Support will ask them to start the wallet, unlock it, then use the "Help > Mark log file" option before trying to add the token that isn't working.

That lets the developer jump immediately to the point in the log that has the information specifically referring to the problem.

sk91:
@dfcowell @rodrigopavezi I propose that we also play with log level while marking the log file.

There is no sense in bloating the log file with `info` level events by default. I think it should always be `warn`. 

But when the user clicks on `Mark log file` we should switch it to `info` to get more details

rhdelima:
Passed QA

itsmeterrylin:
Closing issue as passed QA.

PR #291, opened 2018-07-12
Feat/269 propagate logger

sk91:
related to #269 
- [x] replace 'silly' level with 'trace'
- [x] add sprintf (to avoid serializing huge objects when level is inappropriate)
- [x] replace all loging to use the new logger


src/common/logger.js
```diff
@@ -66,48 +77,57 @@ export class Logger {
...
 	}
 
-	logMessage(level, msg) {
-		if (msg instanceof Error) {
-			msg = this.constructor.errToStr(msg);
-		}
-
-		msg = this.fmtMessage(msg);
+	logMessage(level, msg, args) {
+		if (!this.constructor.checkLevels(level)) return;
```
dfcowell: Is there some reason this wasn't incorporated in the existing filter function?

src/main/controllers/rpc-handler.js
```diff
@@ -751,7 +752,7 @@ module.exports = function(app) {
 									kycprocess,
 									requiredStaticData[i]._id
 								);
-								console.log('THE static data', staticDatas);
+								log.debug('THE static data %s', staticDatas);
```
dfcowell: Are we sure that this doesn't contain any private info? Now that we're working on integrating Sentry as a log transport we need to be very careful when logging variables.

src/main/db.js
```diff
@@ -8,18 +10,18 @@ const init = async () => {
...
-		console.error(e);
+		log.error(e);
 		const backupPath = `${db.connection.filename}.bkp`;
 		// Tear down connections connected to existing file
 		await knex.destroy();
 		try {
 			createBackup(db.connection.filename, backupPath);
 		} catch (backupError) {
-			console.log('Automatic recovery has already been attempted and failed. Aborting.');
+			log.error('Automatic recovery has already been attempted and failed. Aborting. %s', e);
```
dfcowell: Shouldn't `e` here be `backupError`?

src/main/main.js
```diff
@@ -2,7 +2,7 @@
 'use strict';
 import configureStore from 'common/configure-store';
 import { localeUpdate } from 'common/locale/actions';
-import Logger from 'common/logger';
+const { Logger } = require('common/logger');
```
dfcowell: Why not use the `import` syntax here?

package.json
```diff
@@ -99,6 +99,7 @@
     "request": "2.87.0",
     "selfkey-ui": "https://github.com/SelfKeyFoundation/selfkey-ui.git",
     "source-map-support": "^0.5.6",
+    "sprintf-js": "^1.1.1",
```
dfcowell: We're not using any of the advanced formatting that this library gives us. Pretty much everything we're logging can be handed with Javascript template literals. We should be careful about adding dependencies, since eventually we're going to need to audit all of these.

What do you think about switching to JS template literals throughout the app?

src/common/logger.js
sk91: yes, I am applying filter on a formatted string. I want to check levels before, to avoid formatting.

src/main/controllers/rpc-handler.js
sk91: I believe you are partially right. I think we should strongly define log levels and what log levels will be enabled by default for users in deferent modes (normal and issue reproduction). In any case I think debug and trace should be available only to developers so we should not be concerned with security for thous.

src/main/db.js
sk91: correct, fixing

package.json
sk91: I believe js template literals are useful in most cases.
The purpose of this is to delay formatting until level check. So if you have:
`log.debug('debug message %j', superHugeObject)`
and your log level is `warn`, the message will not be formatted, and `superHugeObject` will not have to be serialized as opposed to:
``log.debug(`debug message ${JSON.stringify(superHugeObject)}`);``

src/main/main.js
sk91: replaces `import Logger from 'common/logger';`  with `const { Logger } = require('common/logger');`
will fix for that file

sk91: in general mixing thouis two styles are a very bad practice. Added fixing it for the whole project as subtask for #268

sk91:
blocked until #294 is resolved

sk91:
unblocked

PR #292, opened 2018-07-13
fix: creating tx-history, tx-history paging, sending tx

tmchedlishvili:


src/main/models/wallet-setting.js
```diff
@@ -44,7 +44,7 @@ class WalletSetting extends BaseModel {
 				showDesktopNotifications: { type: 'integer' },
 				previousTransactionCount: { type: 'integer' },
 				txHistoryLastSyncedBlock: { type: 'integer' },
-				airDropCode: { type: 'integer' }
+				airDropCode: { type: ['integer', 'null'] }
```
sk91: @tmchedlishvili you said it should be string, no?

tmchedlishvili: @sk91 For now I don't know it exactly, it should be nullable definitely, so i added it.

PR #305, opened 2018-07-17
fix: log error for connecting to ledger screen

tmchedlishvili:


src/renderer/angular/controllers/commons/dialogs/connecting-to-ledger-controller.js
```diff
@@ -42,7 +42,7 @@ function ConnectingToLedgerController(
 				$rootScope.openChooseLedgerAddressDialog(accounts, ACCOUNTS_QUANTITY_PER_PAGE);
 			})
 			.catch(err => {
-				log.error(err);
+				log.error(err.message);
```
dfcowell: This doesn't log the stacktrace. @sk91 `log.error(err)` should work here, right?

tmchedlishvili: @dfcowell `log.error(err)` does't works, there is an error while logging the `err` object, 
so next line `onError();` newer be executed.
An issue is: connection to ledger screen is newer be disappeared when ledger is not well connected.

tmchedlishvili: @dfcowell Apparently the logging issue is fixed by another commit, so i'm closing this pull request.

PR #313, opened 2018-07-19
Feat/268 refactor id attributes

sk91:
related to #268


sk91:
selfkey-identity-attributes

PR #316, opened 2018-07-20
refactor(services): organize files by modules WIP

sk91:
related to #268


src/blockchain/tx-history-service.js
```diff
@@ -0,0 +1,299 @@
...
+export const ETH_BALANCE_DIVIDER = new BigNumber(10 ** 18);
+export const ENDPOINT_CONFIG = {
+	1: { url: 'https://api.etherscan.io/api' },
+	3: { url: 'http://api-ropsten.etherscan.io/api' }
+};
+export const API_ENDPOINT = ENDPOINT_CONFIG[config.chainId].url;
+
+export let OFFSET = 1000;
+
+export const TX_LIST_ACTION = `?module=account&action=txlist&sort=asc&offset=${OFFSET}`;
```
tmchedlishvili: @sk91 can you update the entire file with current dev version please? 
https://github.com/SelfKeyFoundation/Identity-Wallet/blob/dev/src/main/controllers/tx-history-service.js

sk91:
rebased over dev

src/blockchain/tx-history-service.spec.js
```diff
@@ -0,0 +1,244 @@
...
+			expect(processed).toEqual(processedEthTransaction);
+		});
+		it('should process token transaction', async () => {
+			service = new TxHistoryService(web3ServiceMock);
+			let processed = _.omit(await service.processTx(tokenTransaction), 'createdAt');
+			expect(processed).toEqual(processedTokenTransaction);
+		});
+	});
+	xit('processTxHistory', () => {});
+	xit('getWalletSetting', () => {});
```
dfcowell: Are these TODOs?

src/exchanges/exchanges-service.js
```diff
@@ -0,0 +1,24 @@
...
+
+export class ExchangesService {
+	static async loadExchangeData() {
+		const response = await fetch(`${airtableBaseUrl}Exchanges`);
+
+		const responseBody = await response.json();
+
+		const exchanges = responseBody.Exchanges.filter(
+			row => row.data && row.data.fields.name
+		).map(row => ({
```
dfcowell: The data structure produced here seems a bit weird.

src/logger/logger.js
```diff
@@ -30,16 +30,21 @@ export class Logger {
...
+			args.shift(msg);
+			msg = '%2j';
+		}
 		args = args.map(e => (e instanceof Error ? errToStr(e) : e));
 		if (typeof msg === 'string') {
 			try {
 				msg = vsprintf(msg, args);
 			} catch (error) {
 				this.error(error);
+				this.error('could not format msg ' + msg);
```
dfcowell: We should prefer using template strings where possible.

src/main/main.js
```diff
@@ -1,18 +1,26 @@
 /* global __static */
```
dfcowell: This massive list of imports is a pretty good indicator that this file needs some refactoring in the future.

src/main/main.js
```diff
@@ -82,51 +88,41 @@ let isSecondInstance = electron.app.makeSingleInstance((commandLine, workingDire
...
 
-		const PriceService = require('./controllers/price-service');
-
-		const AirtableService = require('./controllers/airtable-service');
-
-		const LedgerService = require('./controllers/ledger-service')();
-		electron.app.ledgerService = new LedgerService();
+		const priceService = (electron.app.priceService = new PriceService());
 
+		electron.app.web3Service = new Web3Service();
```
dfcowell: We should get away from attaching all of these dependencies to the `electron` object.

src/renderer/angular/services/ledger.service.js
```diff
@@ -0,0 +1,88 @@
...
+const log = new Logger('LedgerService');
+const EthUnits = require('../classes/eth-units');
+const async = require('async');
+
+function LedgerService(RPCService, CommonService, Web3Service) {
+	'ngInject';
+
+	log.info('LedgerService Initialized');
+
+	class LedgerService {
```
dfcowell: We should try to move all of this device interaction logic out of the renderer and into the node process in the future. Having it tightly coupled to Angular and the presentation layer makes it hard to re-use for cases like the browser extension, where we might want to trigger transaction signing flows remotely.

src/exchanges/exchanges-service.js
sk91: how strange? an array of {name: ... , data: ...} I did not modify the schema

src/main/main.js
sk91: @dfcowell we have a plan already with @rodrigopavezi how to manage global dependancies, we'll do it in a separate issue

src/renderer/angular/services/ledger.service.js
sk91: it should not be here, probably a merge error. Fixing

src/main/main.js
sk91: yes, a massive one )

src/logger/logger.js
sk91: fixing

src/blockchain/tx-history-service-old.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
-const { Logger } = require('common/logger');
+const { Logger } = require('logger/logger');
```
rodrigopavezi: @sk91 you know that common is shared between processes right? How will we access logger from the renderer one?

src/renderer/react/price-box/containers/price-box.js
```diff
@@ -1,6 +1,6 @@
 import { connect } from 'react-redux';
 import { CryptoPriceBox } from 'selfkey-ui';
-import { getLocale } from 'common/locale/selectors';
+import { getLocale } from 'locale/selectors';
```
rodrigopavezi: @sk91 here is the same. I think we should keep what is common between process on the common directory. Thoughts?

rodrigopavezi:
@sk91 another thing, can you bump the version to 0.1.0-beta?

src/exchanges/exchanges-service.js
dfcowell: I guess what I'm wondering is why we're doing this processing on the object in the first place. This is what the input/output (with fake field names because I'm not sure of the real ones) looks like:

```
// Input:

{ name: "foo", field1: "value", field2: "value" }

// Output:

{ name: "foo", data: { name: "foo", field1: "value", field2: "value" } }
```

Surely something like:

```
...
).map(row => row.data.fields)
```

would be more convenient than nesting the same data inside this "data" attribute.

Let's keep it with the existing schema for now. We can refactor later.

sk91: that's because it goes to sqlite, so data becomes the json field that stores a generic schema

PR #317, opened 2018-07-21
Feat/sel 86 my crypto

rodrigopavezi:


rodrigopavezi:
@dfcowell @sk91 please review specially last commit where it has price update through redux. Cheers

src/main/controllers/price-service.js
```diff
@@ -61,12 +64,39 @@ const loadPriceData = async () => {
...
+				id: existing[tokenData.short].id,
+				name: tokenData.long,
+				symbol: tokenData.short,
+				source: 'https://coincap.io',
+				priceUSD: +tokenData.price,
+				priceBTC: +tokenData.price / btcPriceUsd,
+				priceETH: +tokenData.price / ethPriceUsd,
+				updatedAt: Date.now()
+			};
+			await TokenPrice.bulkEdit([tokenPrice]);
```
sk91: for one token better use `TokenPrice.updateById(tokenPrice.id, tokenPrice)`

src/main/controllers/price-service.js
```diff
@@ -61,12 +64,39 @@ const loadPriceData = async () => {
...
+
+			emitter.emit('pricesUpdated', await TokenPrice.findAll());
```
sk91: probably need to change pricesUpdated logic to allow sending only one symbole, no sense in sending all of them in each change

src/main/controllers/rpc-handler.js
```diff
@@ -1012,9 +1036,18 @@ module.exports = function(app) {
 	};
 
 	controller.prototype.getWalletTokens = function(event, actionId, actionName, args) {
-		WalletToken.findByWalletId(args.walletId)
-			.then(data => {
-				app.win.webContents.send(RPC_METHOD, actionId, actionName, null, data);
+		Wallet.findById(args.walletId)
+			.then(wallet => {
+				WalletToken.findByWalletId(wallet.id)
+					.then(data => {
```
sk91: wallet tokens can be loaded with eager:
`Wallet.findById(id).eager('tokens')`
Than can be used as:
`wallet.tokens`

src/main/models/wallet.js
```diff
@@ -92,6 +92,10 @@ class Wallet extends BaseModel {
 		return this.findAllWithKeyStoreFile().where({ isSetupFinished: 1 });
 	}
 
+	static findById(id) {
+		return this.query().findOne({ id });
```
sk91: `this.query().findById(id)` would fit better here

src/renderer/angular/directives/commons/sk-token-box.directive.js
```diff
@@ -52,5 +56,12 @@ function SkTokenBoxDirective($rootScope, $window, $timeout, CommonService) {
 		templateUrl: 'common/directives/sk-token-box.html'
 	};
 }
-SkTokenBoxDirective.$inject = ['$rootScope', '$window', '$timeout', 'CommonService'];
+SkTokenBoxDirective.$inject = [
+	'$rootScope',
+	'$log',
```
sk91: you probably accidentally added `$log` on merge

src/renderer/react/my-crypto/containers/crypto-chart-box.js
```diff
@@ -0,0 +1,34 @@
+import { bindActionCreators } from 'redux';
+import { connect } from 'react-redux';
+import * as localeActions from 'common/locale/actions';
+import * as fiatCurrencyActions from 'common/fiatCurrency/actions';
```
sk91: shouldn't this be operations instead of actions?

src/main/controllers/rpc-handler.js
rodrigopavezi: @sk91 I tried that but it is not returning the tokens and instead undefined. I think we can go back to it when we refactor the token persistence. Thoughts? Cheers

sk91: @rodrigopavezi strange, ok will check it out in a separate pull request

sk91:
Looks good!

PR #323, opened 2018-07-26
Feat/sel 264 trezor integration

tmchedlishvili:


src/main/controllers/trezor-service.js
```diff
@@ -0,0 +1,168 @@
...
+
+const hardeningConstant = 0x80000000;
+const defaultAddress = [
+	(44 | hardeningConstant) >>> 0,
+	(60 | hardeningConstant) >>> 0,
+	(0 | hardeningConstant) >>> 0,
+	0
+];
+
+const getAddressByIdex = index => {
```
rodrigopavezi: @tmchedlishvili is this a typo and should be `getAddressByIndex`?

src/main/controllers/trezor-service.js
```diff
@@ -0,0 +1,168 @@
...
+	return defaultAddress.concat([+index]);
+};
+
+const addHexPrefix = val => {
```
rodrigopavezi: @tmchedlishvili I think I have seem similar function somewhere. Couldn't we reuse it if that is the case?

src/main/controllers/trezor-service.js
```diff
@@ -0,0 +1,168 @@
...
+	}
+	function pinCallback(type, callback) {
+		_emitter.on('ON_PIN', (err, enteredPin) => {
+			callback(err, enteredPin);
+		});
+
+		_emitter.emit('TREZOR_PIN_REQUEST');
+	}
+
+	async function getCurrentSession() {
```
rodrigopavezi: should it have a underscore (_) as prefix because it is a private function?

src/main/controllers/trezor-service.js
```diff
@@ -0,0 +1,168 @@
...
+			txData.gasLimit,
+			txData.to,
+			txData.value,
+			txData.data,
+			chainId
+		);
+
+		let signed = null;
+		try {
+			signed = await timeout(signPromise, 30000);
```
rodrigopavezi: can you move this timeout to a config or at least a const in the top of this service class?

src/renderer/angular/services/hardware-wallet.service.js
```diff
@@ -0,0 +1,85 @@
...
+const log = new Logger('HardwareWalletService');
+const EthUnits = require('../classes/eth-units');
+
+function HardwareWalletService($rootScope, RPCService, CommonService, Web3Service) {
+	'ngInject';
+
+	log.info('HardwareWalletService Initialized');
+	RPCService.on('TREZOR_PIN_REQUEST', () => {
+		$rootScope.openEnterTrezorPinDialog();
+		console.log('mevida1111111111111111');
```
rodrigopavezi: I think we should not leave console.log here.

PR #338, opened 2018-07-29
feat: react component for transfer price

rodrigopavezi:


src/renderer/angular/controllers/member/wallet/manage-token-controller.js
```diff
@@ -15,8 +15,8 @@ function ManageTokenController(
 	log.debug('ManageTokenController, %j', $stateParams);
 
 	let temporaryMap = {
-		key: 'Selfkey',
-		eth: 'Ethereum'
+		KEY: 'Selfkey',
```
tmchedlishvili: @rodrigopavezi 
should it be "SelfKey", with uppercase "K" ?

dfcowell: Yes, once this typo is fixed we can merge this @rodrigopavezi.

Issue #350, opened 2018-07-31
visibility icons on manage token list is not shown


sk91:
#351

itsmeterrylin:
Closing issue as passed QA.

PR #351, opened 2018-07-31
fix(price-table): visibility buttons

sk91:
#350

### DIFF
--- a/src/renderer/react/my-crypto/containers/crypto-price-table.js
+++ b/src/renderer/react/my-crypto/containers/crypto-price-table.js
@@ -9,7 +9,8 @@ const mapStateToProps = state => {
 	return {
 		...getLocale(state),
 		...getFiatCurrency(state),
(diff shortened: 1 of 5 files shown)


src/renderer/react/my-crypto/containers/crypto-price-table.js
```diff
@@ -9,7 +9,8 @@ const mapStateToProps = state => {
 	return {
 		...getLocale(state),
 		...getFiatCurrency(state),
-		tokens: getTokens(state)
+		tokens: getTokens(state),
+		alwaysVisible: ['', '0x4cc19356f2d37338b9802aa8e8fc58b0373296e7']
```
rodrigopavezi: @sk91 what about Ethereum? Will it use the same address?

sk91: @rodrigopavezi Ethereum address is '' -> no address

Issue #330, opened 2018-07-27
Move from Travis to Circle Ci


rodrigopavezi:
Circle ci has proven to be better than Travis with its Circle Ci 2.0 features where we can run separated jobs for each phase of the build. If tests fails we won't need to re run the whole build for instance. It will require only to run the test job. That should improve our development and also make us to have one less CI tool to maintain.

PR #357, opened 2018-08-01
Feat/cci for mac

rodrigopavezi:
Improvement #330


.circleci/config.yml
```diff
@@ -1,104 +1,267 @@
...
-            - yarn-packages-{{ .Branch }}-{{ checksum "yarn.lock" }}
-            - yarn-packages-{{ .Branch }}
-            - yarn-packages-master
-            - yarn-packages-
-      - run:
-          name: E2E Tests
-          command: OSENV=circle yarn test
+    'unit-test':
+        docker: &DOCKERIMAGE
+            - image: circleci/node:9-stretch-browsers
```
sk91: node versions should be aligned with .nvmrc

.circleci/config.yml
```diff
@@ -1,104 +1,267 @@
...
+                      curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+                      sudo apt-get update && sudo apt-get install google-cloud-sdk jq
+            - run:
+                  name: Deploy mac to staging
+                  command: bash deploy.sh "$(git log --format=%B -n 1 $CIRCLE_SHA1)"
+                  environment:
+                      OSENV: mac
 
+    'stage-linux':
+        docker: *DOCKERIMAGE
```
dfcowell: Since we're using the docker executor here, we could just use the Cloud SDK docker image (https://hub.docker.com/r/google/cloud-sdk/) here instead of manually installing it.

Issue #265, opened 2018-07-09
“Not Enough Gas": Update Transaction Message Display & UX


dfcowell:
The error message that appears when the wallet user does not have enough gas to send tokens is incorrect and unhelpful. The phrasing of the message may also cause distress to a user unfamiliar with how ERC-20 tokens work.

Wireframe: https://balsamiq.cloud/sv8gyk0/piks2o2/r0329

Key Functionality:
1) User can get notified in a friendly way that they don't have enough gas (ETH) to pay for this transaction. 
2) Display this message across ETH, KEY, and ERC-20 sending transaction workflows.

Dependency:
Please prioritize Issue #312 first as the displayed network transaction fee will be pulled from ETH Gas Station. We can then compare this to the current wallet balance and trigger this message earlier in the workflow for a better UX.

itsmeterrylin:
@rodrigopavezi @rhdelima  For URL to 'here' below:

https://help.selfkey.org/article/87-how-does-gas-impact-transaction-speed

sk91:
@itsmeterrylin @rodrigopavezi  fixed in #379 awaiting review

rhdelima:
Passed on MAC. Waiting for a WIN build

rhdelima:
Passed QA on WIN

PR #369, opened 2018-08-07
fix: not enough gas error message

rodrigopavezi:
#265


src/renderer/templates/common/dialogs/send-token.pug
```diff
@@ -1,5 +1,7 @@
 .send-token
-    .send-token-wrapper
+    div(style="width:600px;" ng-if="errors.sendFailed")
```
sk91: can this trigger on a different error than "has no gas"?

PR #373, opened 2018-08-13
feat(di): introduce di and use it for services

### LINKED ISSUE #364: dependency management of singelton services

sk91:
resolves #364

### DIFF
--- a/src/main/main.js
+++ b/src/main/main.js
@@ -13,7 +13,7 @@ import { getUserDataPath, isDevMode, isDebugMode, isTestMode } from 'common/util
 import config from 'common/config';
 import createMenuTemplate from './menu';
 import db from './db/db';
-import { configureContrainer } from './composition-root';
+import { configureContext } from './context';
 
 const log = new Logger('main');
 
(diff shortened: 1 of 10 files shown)


sk91:
solution inspired by [https://medium.com/@Jeffijoe/dependency-injection-in-node-js-2016-edition-f2a88efdd427](this post)

using the [awilix](https://github.com/jeffijoe/awilix) di container

src/main/main.js
```diff
@@ -99,30 +93,22 @@ function onReady(app) {
 			appUpdater();
 		}
 		await db.init();
-		if (!isDevMode() && !isTestMode()) {
-			await crashReportService.startCrashReport();
-		}
 		const store = configureStore(global.state, 'main');
+		const container = configureContrainer(store, app);
+		const di = container.cradle;
```
rodrigopavezi: @sk91 would make more sense to call this `applicationContext` instead of di? Thoughts?

rodrigopavezi: Also, there is a typo on method name and I think it should be `configureContainer`. Or if we are calling applicationContext we could call it `configureContext`. Thoughts?

rodrigopavezi:
looks good 👍

Issue #365, opened 2018-08-06
JSON Schema: Update Identity Attribute Forms


sk91:
https://mozilla-services.github.io/react-jsonschema-form/

Aha! Link: https://kycc.aha.io/features/SKEY-199

PR #377, opened 2018-08-14
feat(json-schema): json-schema-forms mvp

sk91:
implement json-schema-forms in identity wallet
related to #365


sk91:
For each field type from air-table we check in appropriate place on `platform.selfkey.org`. If schema is found, it is derefernced and cached in schemas table (for 5h).

In renderer, json-form rendering will be triggered only if there is a resolved schema, otherwise the old style form will be used for this field type.

This pull request has the following issues that need to be resolved separately:
- migrate json-schema-form handling to selfkey-ui repo and implement custom ui elements for it
- handle [json-ui-schemas](https://github.com/mozilla-services/react-jsonschema-form#form-customization)
- styles included are very minimal, each time a new schema is uploaded to platform.selfket.org we should check if it is styled correctly

src/main/identity/id-attribute-type-service.js
```diff
@@ -1,20 +1,57 @@
...
+				type: type.key
+			};
+			if (type.schema) {
+				return type.schema.$query().patch(schema);
+			}
+			return IdAttributeSchema.query().insertAndFetch(schema);
+		} catch (error) {
+			log.error(error);
+		}
+	}
```
rodrigopavezi: @sk91 wondering, shouldn't this be part of a schema class so other classes can extend it? Cheers

src/main/identity/id-attribute-type-service.js
```diff
@@ -1,20 +1,57 @@
...
 import fetch from 'node-fetch';
+import { Logger } from 'common/logger';
+import RefParser from 'json-schema-ref-parser';
 import IdAttributeType from './id-attribute-type';
+import { IdAttributeSchema } from './id-attribute-schema';
 
+const log = new Logger('id-attribute-type-service');
 const airtableBaseUrl = 'https://alpha.selfkey.org/marketplace/i/api/';
-
+export const JSON_SCHEMA_EXPIRTION_TIME = 1000 * 60 * 60 * 5; // 5h
```
rodrigopavezi: @sk91 I think we should use the same expiration time for all schemas and store it in a global configuration class. Thoughts? Cheers

src/main/identity/id-attribute-type-service.js
```diff
@@ -1,20 +1,57 @@
...
+export const JSON_SCHEMA_URL = 'http://platform.selfkey.org/schema/attribute';
```
rodrigopavezi: @sk91 I would also move `http://platform.selfkey.org/schema/` to a global config and just append `attribute` to it. Thoughts?

dfcowell: Something to remember: not all schemas will start with `http://platform.selfkey.org/schema/`

One of the reasons we're going down this path is so that third parties can define ID attributes for the wallet.

I don't even think this should be a configuration value. We should seed a schema list somewhere (whether a cache table or some other data store) with a list of our standard schemas, but they should be stored as fully-qualified URLs - otherwise we need to add special handling for our own schemas, which isn't a great design decision.

sk91: @rodrigopavezi @dfcowell that's exactly why schema has "jsonSchemaUrl" property. JSON_SCHEMA_URL is used only if it is not provided. 

So @rodrigopavezi no, I don't think we should have it as configuration, it will be refactored away in one of the next pull requests (when we start receiving schema url together with id-attribute-type)

@dfcowell @ben has committed a json list of schemas to platforms if I remember correctly. In next iterations, we will start using it instead of the air-table

src/main/identity/id-attribute-type-service.js
sk91: @rodrigopavezi no, because this is the only schema class we have right now, the other one is the model, which in our case used only to interact with the database

src/main/identity/id-attribute-type-service.js
sk91: I think this value,  same as the schemaUrl value below, should be defined separately as schema properties. Here we have one value for all but it is only for mvp purposes>

in general schema expiration should be different for each schema

src/main/identity/id-attribute-type-service.js
rodrigopavezi: @sk91 @dfcowell I see now. makes sense. cheers

Issue #368, opened 2018-08-07
fix tests for all platforms


sk91:
- [x] fix disabled unit tests on mac #371
- [x] fix unit tests on windows #378 
- [x] fix e2e tests on windows #378 
- [x] fix unit tests on linux #378 
- [x] fix e2etests on linux #378 
- [x] fix sporadic unit test failure

PR #378, opened 2018-08-16
Fix/tests

sk91:
related to #368


src/main/db/test-db.js
```diff
@@ -1,33 +1,56 @@
...
-		await db.knex.raw('select 1+1 as result');
-		return false;
-	} catch (err) {
-		await db.knex.client.initializePool(db.config);
-		return true;
+export class TestDb {
+	static async init() {
+		try {
+			this.knex = Knex(config);
+			this.oldKnex = Model.knex();
```
dfcowell: Why are we keeping a reference to the old instance of knex?

src/main/db/test-db.js
```diff
@@ -1,33 +1,56 @@
...
+		}
 	}
-	await init();
-};
-
-export { init, reset };
+	static async reset() {
+		try {
+			await this.knex.destroy();
+			Model.knex(this.oldKnex);
```
dfcowell: I don't understand what this is doing.

sk91: @dfcowell that's a result of a very long fight with db connection errors in tests. This one is a leftover, will remove

sk91:
@dfcowell fixed

package.json
```diff
@@ -32,12 +32,14 @@
   "main": "./src/main/main.js",
   "scripts": {
     "dev": "gulp && electron-webpack dev",
-    "postinstall": "electron-builder install-app-deps buildDependenciesFromSource",
+    "install-app-deps": "electron-builder install-app-deps buildDependenciesFromSource",
+	"install-all": "yarn && yarn install-app-deps",
```
rodrigopavezi: @sk91 curious, why did you remove `postinstall`?  Cheers

rodrigopavezi:
@sk91 looks really good just a comment about the install-all. cheers

package.json
sk91: @rodrigopavezi it takes a lot of time becouse it rebuilds everything from source. For unit tests it is not needed

rodrigopavezi: @sk91 makes sense. cheers

PR #389, opened 2018-08-21
fix(eslint): fix eslint on commit

sk91:


sk91:
Changes already in dev

PR #390, opened 2018-08-22
Feat/integrate lws

sk91:


sk91:
@altninja yes, basically that's it

PR #430, opened 2018-09-07
fix(lws): auth attributes

sk91:


src/main/lws/lws-service.js
```diff
@@ -156,14 +164,46 @@ export class LWSService {
...
+				check.privateKey
+			);
+
+			if (!signature) {
+				return conn.send(
+					{
+						error: true,
+						payload: {
+							code: 'sign_error',
+							message: 'Cannot could not generate signature'
```
altninja: double negative - select one of either cannot or could not

src/main/lws/lws-service.js
```diff
@@ -131,14 +131,22 @@ export class LWSService {
...
+			const signature = ethUtil.ecsign(msgHash, Buffer.from(privateKey, 'hex'));
+			return signature;
+		} catch (error) {
+			log.error(error);
 			return null;
 		}
-		const msgHash = ethUtil.hashPersonalMessage(Buffer.from(nonce, 'hex'));
-		const signature = ethUtil.ecsign(msgHash, Buffer.from(check.privateKey, 'hex'));
-		return signature;
+	}
```
altninja: is this needed?  In the implementation I have on local I'm able to just do JSON.stringify to the original return signature result and it works correctly with the verification on the server side

src/main/lws/lws-service.js
```diff
@@ -156,14 +164,46 @@ export class LWSService {
...
 	async reqAuth(msg, conn) {
 		try {
+			let check = this.checkWallet(msg.payload.publicKey, conn);
+			if (!check.unlocked) {
+				return conn.send(
+					{
+						error: true,
+						payload: {
+							code: 'auth_error',
+							message: 'Cannot auth with locked wallet'
```
altninja: are these messages for display in UI as well?

src/main/lws/lws-service.js
```diff
@@ -156,14 +164,46 @@ export class LWSService {
...
 			const nonceResp = await this.fetchNonce(msg.payload.website.apiUrl);
-			if (nonceResp.error) {
+			if (nonceResp.error || !nonceResp.nonce) {
 				return conn.send(
 					{
 						error: true,
 						payload: {
 							code: 'nonce_fetch_error',
-							message: nonceResp.error
+							message: nonceResp.error || 'No nonce in responce'
```
altninja: typo: response

src/main/lws/lws-service.js
```diff
@@ -172,16 +212,18 @@ export class LWSService {
 			const body = {
 				publicKey: msg.payload.publicKey,
 				nonce: nonceResp.nonce,
-				signature: this.genSignature(nonceResp.nonce, msg.payload.publicKey, conn)
+				signature: Buffer.from(this.stringifySignature(signature), 'utf8').toString(
```
altninja: we should include this together with the signature generation, modification functions and then just return the final result to the request body object

src/main/lws/lws-service.js
sk91: Yes, you are transferring stringified Buffer, it will work only between node projects. We don't know which server tech our customers will use.

src/main/lws/lws-service.js
sk91: Right now - no, but we can make them

sk91:
@altninja Fixed staff

PR #433, opened 2018-09-10
feat: stacking part 1 show no balance

rodrigopavezi:
@sk91 this is still a working in progress as we cannot merge to dev. We need the full stacking for that. But it is good to separate in small chunks for make it easier for review. Cheers


package.json
```diff
@@ -117,9 +117,10 @@
     "request": "2.87.0",
     "reselect": "^3.0.1",
     "scrypt": "^6.0.3",
-    "selfkey-ui": "https://github.com/SelfKeyFoundation/selfkey-ui.git",
+    "selfkey-ui": "https://github.com/SelfKeyFoundation/selfkey-ui.git#feat/stack-1",
```
sk91: using hashes instead of branches works great after merge, better to use them

src/common/exchanges/reducers.js
```diff
@@ -0,0 +1,20 @@
...
+	list: []
+};
+
+const exchangesReducer = (state = initialState, action) => {
+	switch (action.type) {
+		case types.EXCHANGES_UPDATE:
+			console.log('Exchanges', action);
+			return {
+				...state,
+				list: action.payload
```
sk91: list should store ids/keys only with byId near it to normalize state

package.json
rodrigopavezi: I know this is just for now that is why it is merging to staging/stacking. It will be changed when merging to dev

sk91:
Need to do something with the coverage though ...

rodrigopavezi:
@sk91 yeap, I will add tests for the redux stuff so our coverage continues to increase.

PR #435, opened 2018-09-11
feat(tx-history): ui improvements

tmchedlishvili:


src/renderer/angular/directives/commons/sk-tx-history.directive.js
```diff
@@ -1,7 +1,8 @@
 'use strict';
+/* eslint-disable */
```
sk91: why disable enlist ?

tmchedlishvili: @sk91 sorry about this, the line is removed now.

sk91:
@tmchedlishvili If you'll fin a way to eliminate the coverage decrease will be great

PR #437, opened 2018-09-12
fix(376): update Tos, remove a text from marketplace page

tmchedlishvili:


src/renderer/templates/member/marketplace/exchange-item.pug
```diff
@@ -8,7 +8,6 @@
                     img.logo-image(ng-src="{{realData.logo[0].thumbnails.small.url}}", alt="")
                     span.name(style="font-weight: 600") {{realData.name}} (Exchange)
                 .status(layout="row")
-                    span(style='color: #fff') Your Staking Status
```
rodrigopavezi: @tmchedlishvili can you confirm with @itsmeterrylin if should the whole status be removed, not just the text? Cheers

PR #447, opened 2018-09-18
Staking eth

sk91:
to run staking integration test in dev start app via:
`PRIMARY_TOKEN_OVERRIDE=ki LOG_FILTER=MAIN\ main,XXX ENABLE_STAKING_TEST=1 yarn dev`

## usage:

see `main.js -> startStakingTest` for usage examples

`stakingService` is now available in `ctx`

`options`: 
 - for simple transaction: `{from: walletAddress}`
- for gas estimation of a call: `{from:walletAddress, method: 'estimateGas'}`
- to set gas for transactions: `{from: walletAddress, gas: gasLimit, gasPrice: gasPrice}` if not set will be resolved automatically

### `async acquireContract() `

- called automatically in app startup
- fetches data from airtable
- initializes `stakingService.activeContracts`
- initializes `stakingService.deprecatedContracts` (array)
- iniitalizes `stakingService.tokenContract` 

### `async placeStake(ammount, serviceAddress, serviceId, options)` 

- resolves an object `{approve: approveHash, deposit: depositHash}` 
- place stake with estimateGas will return same object with gas etimation instead of hashes
- if user already has enough allowance to make the deposit, `approve` key will not be set

### `async getStakingInfo(serviceAddress, serviceId, options)`

- resolves object in the form of ` { balance: 0, serviceAddress, serviceId, contract: null, releaseDate: 0 };`
- if staking is not found in active . contract, deprecated contracts will be checked
- contract key will point to the contract that has the stake
- if contract with stake is deprecated, releaseDate will always be 0
- if call to contract info has error, `balance` will be set to 0

### `async withdrawStake(serviceAddress, serviceId, options)`
- resolves a hash with withdraw
- calls to `getStakingInfo` and throws error if no stake or stake is locked

## `StakingContract`
- available after `acquireContract`
- available via `stakingService.activeContract`
- available via `stakingService.deprecatedContracts`

### `getBalance(serviceAddress, serviceId, options)`
- resolves staking balance

### `deposit(ammount, serviceAddress, serviceId, options) `

- resolves deposit hash
- on gasEstimate will resolve gas

### `withdraw(serviceAddress, serviceId, options)`

- resolves withdraw hash
- on gasEstimate will resolve gas

### `getReleaseDate(serviceAddress, serviceId, options) `

- resolves release date ts

### `getLockPeriod(serviceAddress, serviceId, options)`

- resolves lock period

## `TokenContract`
- available after `acquireContract`
- available via `stakingService.tokenContract`

### `approve(depositVaultAddress, maxAmmount, options)`

- resolves approve hash
- on gasEstimate will resolve gas

### `allowance(depositVaultAddress, options)`

- resolves current allowance in token contract


src/main/blockchain/staking-service.js
```diff
@@ -0,0 +1,259 @@
+import fetch from 'node-fetch';
+
+import CONFIG from 'common/config';
+import { abi as SELFKEY_ABI } from 'main/assets/data/abi.json';
+import { Token } from '../token/token';
+import BN from 'bignumber.js';
+// import { TxHistory } from './tx-history';
```
rodrigopavezi: @sk91 I think we should not leave comment code like that.

src/main/blockchain/staking-service.js
```diff
@@ -0,0 +1,259 @@
...
+	constructor({ web3Service }) {
+		this.activeContract = null;
+		this.deprecatedContracts = [];
+		this.web3 = web3Service;
+	}
+	async getStakingInfo(serviceAddress, serviceId, options) {
+		let info = { balance: 0, serviceAddress, serviceId, contract: null, releaseDate: 0 };
+		let contracts = [this.activeContract].concat(this.deprecatedContracts);
+		let balance = 0;
+		options = { ...options };
```
rodrigopavezi: @sk91 why are you doing this with options?

src/main/blockchain/staking-service.js
```diff
@@ -0,0 +1,259 @@
...
+		for (let i = 0; i < contracts.length; i++) {
```
rodrigopavezi: @sk91 curious why didn't you use forEach here? You could use it with a return for continuing the loop. No?

src/main/main.js
```diff
@@ -343,3 +348,68 @@ function handleSquirrelEvent() {
 	}
 	log.info('end handleSquirrelEvent');
 }
+
+function startStakingTest(ctx) {
```
rodrigopavezi: @sk91 curious, why not run this test with Jest?

sk91: @rodrigopavezi it is an integration test, not unit test. Only for dev purposes right now.
Test like this will be run in the new blockchain - integration test suite that is currently WIP.

src/main/blockchain/staking-service.js
sk91: It is the same as doing "clone" on options. It is a general good practice to make sure it will not change the original 'options' object

src/main/blockchain/staking-service.js
sk91: No reason really, is there any objection agains using a loop?

src/main/blockchain/staking-service.js
sk91: right missed it

src/main/blockchain/staking-service.js
sk91: @rodrigopavezi sorry, just have seen that I am breaking the loop as soon as I find a suitable contract. This is not a good usecase for `forEach`

src/main/blockchain/staking-service.js
sk91: @rodrigopavezi fixed

src/main/blockchain/staking-service.js
```diff
@@ -0,0 +1,258 @@
...
+			info.balance = balance;
+			info.contract = contracts[i];
+			if (!contracts[i].isDeprecated) {
+				info.releaseDate = await contracts[i].getReleaseDate(
+					serviceAddress,
+					serviceId,
+					options
+				);
+			}
+			return info;
```
sk91: here @rodrigopavezi. this breaks the `forEach` approach

PR #451, opened 2018-09-24
fix(balance update): update balance ofter sending transaction

tmchedlishvili:


src/main/rpc-handler.js
```diff
@@ -961,11 +961,12 @@ module.exports = function(cradle) {
 			});
 	};
 
-	controller.prototype.getWalletByPublicKey = function(event, actionId, actionName, args) {
+	controller.prototype.getWalletByPublicKey = async function(event, actionId, actionName, args) {
```
sk91: the function should use `await` instead of promise api

src/renderer/angular/services/balance-updater.service.js
```diff
@@ -0,0 +1,62 @@
...
+		let walletUpdaterPromise = RPCService.makeCall('getWalletByPublicKey', {
+			publicKey: $rootScope.wallet.getPublicKey()
+		});
+		let tokensUpdaterPromise = RPCService.makeCall('getWalletTokens', {
+			walletId: $rootScope.wallet.id
+		});
+
+		await Promise.all([walletUpdaterPromise, tokensUpdaterPromise]);
+		let currentWallet = await RPCService.makeCall('getCurrentWallet');
+		console.log(oldBalance, currentWallet.balance);
```
sk91: why not use store directly here with a getCurrentWallet selector? (we want to deprecate RPCService in near future)

tmchedlishvili:
@rodrigopavezi 
the requested changes are made, please review the PR again.

rodrigopavezi:
@tmchedlishvili merging this as we will need to change things for the new advance transaction. cheers

PR #509, opened 2018-10-21
Feat/address book

rodrigopavezi:


src/common/address-book/actions.js
```diff
@@ -0,0 +1,66 @@
+import * as types from './types';
+import { createAliasedAction } from 'electron-redux';
+import { getGlobalContext } from '../context';
+
+const addEntry = async entry => {
```
sk91: - async action should be an operation (like any non atomic action). Atomic actions always update redux store in a predictable way.
- shouldn't async action return a thunk like this?:
``` entry => async (dispatch, getState) => { ```

File: src/common/address-book/operations.js [hunk 2]
```diff
@@ -0,0 +1,90 @@
+import * as actions from './actions';
+import { walletSelectors } from '../wallet';
+import { getAddresses } from './selectors';
+import { getGlobalContext } from '../context';
+
+const init = () => async (dispatch, getState) => {
```
sk91: maybe better to call this loadAddressBook  or something like that?

src/common/address-book/operations.js
```diff
@@ -0,0 +1,90 @@
...
+	await dispatch(actions.loadAddressBookEntries(walletSelectors.getWallet(getState()).id));
+};
+
+const initAdd = () => async dispatch => {
```
sk91: resetAddForm?

src/common/address-book/selectors.js
```diff
@@ -0,0 +1,18 @@
...
+
+export const getAddressError = state => {
+	return state.addressBook.addressError;
+};
+
+export const getLabel = (state, id) => {
+	const entry = state.addressBook.entries.filter(entry => {
+		return entry.id === id;
+	});
+	return entry[0].label;
```
sk91: it will fail if no entry with this id

src/common/address-book/types.js
```diff
@@ -0,0 +1,17 @@
+const ADD_ADDRESS_BOOK_ENTRY = 'app/addressBook/ADD';
```
sk91: type should be: `${SUBJECT}_${ACTION}`. In this case ADRESS_BOOK_ENTRY_ADD.
@rodrigopavezi  we discussed it, but we also need to document it somewhere

test/utils/tools.js
```diff
@@ -39,8 +39,9 @@ function appStart() {
 // }
 
 function appStop() {
-	if (this.app && this.app.isRunning()) {
-		return this.app.stop();
+	if (app && app.isRunning()) {
+		app.stop();
+		return app.stop();
```
sk91: why twice?

src/common/address-book/actions.js
rodrigopavezi: @sk91 I agree with you but here it is a different case. We need this to run on the main process otherwise it will error out on knex. I tried doing the other around but if I run this code on operations it will error out. Cheers

test/utils/tools.js
rodrigopavezi: @sk91 first time close is called it will show the confirmation screen. So we need to call it again so it will be closed and test will finish smoothly. That should also solve the problem with the windows build. Cheers

sk91: 👍

PR #566, opened 2018-11-08
Json schema attributes

sk91:


sk91:
@rodrigopavezi This will break selfkey-id view. Until we fix it in **https://github.com/SelfKeyFoundation/Identity-Wallet/pull/574** selenium tests will faiil

PR #582, opened 2018-12-04
fix(build): by removing installation of windows-build-tools

rodrigopavezi:


gulpfile.js
```diff
@@ -23,12 +24,15 @@ gulp.task('templates', cb => {
 		})
 		.pipe(
 			tap(function(file) {
+				let fileBase = '';
+				if (os.platform() === 'win32') {
+					// eslint-disable-next-line no-useless-escape
+					fileBase = file.path.replace(file.base, '').replace(/\\/g, '/');
+				} else {
+					fileBase = file.path.replace(file.base, '');
+				}
```
sk91: Should it be like below ?
```
let fileBase = file.path.replace(file.base, '');
if (os.platform() === 'win32') {
    fileBase = fileBase.replace(/\\/g, '/');
}
```

PR #616, opened 2019-02-04
Feature/fully react

rodrigopavezi:


src/common/transaction/actions.js
```diff
@@ -1,10 +1,10 @@
 import * as types from './types';
 import { createAliasedAction } from 'electron-redux';
-import TrezorService from 'main/blockchain/trezor-service';
+// import TrezorService from 'main/blockchain/trezor-service';
```
sk91: should clean comments

package.json
```diff
@@ -142,8 +141,8 @@
 		"striptags": "3.1.1",
 		"sudo-prompt": "8.2.3",
 		"tcp-port-used": "1.0.1",
-		"trezor.js": "6.19.0",
 		"url-join": "^4.0.0",
+		"trezor-wallet-provider": "https://github.com/SelfKeyFoundation/trezor-wallet-provider.git#0c5ef04",
```
sk91: should we publish to npm?

src/main/blockchain/TrezorWallet.js
```diff
@@ -0,0 +1,180 @@
+// import EthereumTx from 'ethereumjs-tx';
+import { publicToAddress } from 'ethereumjs-util';
+// import BigNumber from 'bignumber.js';
```
sk91: clean commnets

src/renderer/index.js
```diff
@@ -79,6 +82,9 @@ render(
 						<Route path="/backupPrivateKey" component={BackupPK} />
 						<Route path="/main" component={Main} />
 						<Route path="/unlockWallet" component={Unlock} />
+						<Route path="/enterTrezorPin" component={EnterPin} />
```
sk91: please use lowercase letters with dashes, it is a url after all

src/renderer/wallet/unlock/ledger/connecting.jsx
```diff
@@ -6,105 +6,204 @@ import {
...
 	}
 }
 
 const mapStateToProps = (state, props) => {
-	console.log('STATE', state);
-	return {};
+	const app = appSelectors.selectApp(state);
+	return {
+		error: app.error,
+		hasConnected: app.hardwareWallets.length > 0
```
sk91: hasConnected should be a separate selector

src/renderer/index.js
rodrigopavezi: @sk91 it is not a web app so there is no real need to be a dashed separated string. We could but at this point there are a lot of routings using this format so I would not change that right now. Perhaps we can change all together later if needed. cheers

PR #617, opened 2019-02-04
Incorporations marketplace

andregoncalves:
A bit of a gigantic pull request.
Should not affect the rest of the code in dev.


src/renderer/angular/controllers/member/marketplace/marketplace-incorporations-controller.js
```diff
@@ -0,0 +1,4 @@
+'use strict';
```
sk91: I believe this and all angular dir should go, right ?

andregoncalves:
@sk91 Angular is not needed, removed. Also removed unneeded imported comments.

PR #624, opened 2019-02-05
feat(address-book): create add, edit, and delete features

eduardosmoura:


src/renderer/address-book/add/index.jsx
```diff
@@ -1,37 +1,188 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-import { AddressBookAdd, ModalBox } from 'selfkey-ui';
+import { StyledButton, ModalBox } from 'selfkey-ui';
 import { addressBookSelectors, addressBookOperations } from 'common/address-book';
+import { Grid } from '@material-ui/core';
+import { withStyles } from '@material-ui/core/styles';
+
+const styles = theme => ({
+	errorText: {
```
rodrigopavezi: @eduardosmoura we will need to remove this css because the material UI theme we created will already have it. For instance for the error you can achieve it like this https://github.com/SelfKeyFoundation/Identity-Wallet/blob/dev/src/renderer/wallet/unlock/private-key.jsx#L122. As we are really close to the deadline I suggest we leave it for later and @designhorf can fix this pretty quick. Thoughts?

eduardosmoura: Sorry, @rodrigopavezi. I didn't know about that. I think @designhorf is more agile than me resolving those CSS issues. Also, I showed him a black space at the of the address page main page. We will fix that later as well. Please, @designhorf let us know your thoughts on this matter? Thank you.

src/renderer/address-book/icon/index.jsx
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react';
```
rodrigopavezi: @eduardosmoura forgot to tell you but we leaving the icons on selfkey-ui so it can be reused. So you need to export it here https://github.com/SelfKeyFoundation/selfkey-ui/blob/master/src/index.tsx#L96 to be able to import in the wallet. Cheers

Issue #607, opened 2019-01-28
Apply for incorporation


rodrigopavezi:
Feature: Apply to Incorporations Program

Background: Given user has opened SelfKey wallet
And the user has opened the Incorporations Marketplace Screen
And the user has opened a specific Incorporations Program

Scenario: Initial application without KYC requirements
Given incorporations detailed program screen is opened
And not all of the KYC requirements are present in the wallet
When user clicks the "Start Incorporation" button
Then ...

Scenario: Existing ongoing application process
Given incorporations detailed program screen is opened
And an existing application process exists for that program
Then the "Start incorporations" button should be deactivated
And a hyperlink to the kyc-chain on-boarding app should be displayed

Scenario: Not enough tokens to apply
Given incorporations detailed program screen is opened
And the user doesn't have enough tokens to cover the program cost
Then the "Start incorporations" button should be deactivated
And message indicating the user doesn't have enough tokens should be displayed

Scenario: Initial application with KYC requirements
Given incorporations detailed program screen is opened
And all of the KYC required attributes and documents are present in the wallet
When user clicks the "Start Incorporation" button
Then a confirmation and authorization screen is displayed (s702)

Scenario: Ask for payment with KEY tokens
Given incorporations detailed program screen is opened
And user clicked the "Start Incorporation" button
And the Confirmation and Authorization screen is shown (s702)
When user clicks "Authorize"
Then a payment confirmation screen should show up

Scenario: Successful payment for incorporations
Given incorporations detailed program screen is opened
And payment confirmation screen is showing up
When user clicks "Pay"
And payment is successful
Then a visual feedback is displayed
And And a hyperlink to kyc-chain on-boarding app is displayed

Scenario: Failed payment for incorporations
Given incorporations detailed program screen is opened
And payment confirmation screen is showing up
When user clicks "Pay"
And payment fails
Then a visual indication is displayed
And ...

rodrigopavezi:
What needs to be implemented:
- [ ] Price on the incorporation item details - [invision](https://projects.invisionapp.com/d/#/console/15222689/319762012/preview)
- [ ] Pay incorporation checkout  - [invision](https://projects.invisionapp.com/d/#/console/15222689/345775295/preview)
- [ ] Incorporation Payment Confirmation - [invision](https://projects.invisionapp.com/d/#/console/15222689/345775294/preview)
[...]

PR #643, opened 2019-02-11
feat: #607 - price on the incorporation item details

andregoncalves:
screen shot 2019-02-11 at 17 30 28

Price on the incorporation item details (invision 701)


andregoncalves:
@rodrigopavezi don't review it jut yet, I just noticed that the KEY value is not properly formatted.

src/renderer/marketplace/incorporation/detail-view/index.jsx
```diff
@@ -1,6 +1,8 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import injectSheet from 'react-jss';
```
rodrigopavezi: @andregoncalves is there a need to react-jss here?

andregoncalves: Nice catch @rodrigopavezi , some old incorporations code still used react-jss.
I ended up refactoring and making the same change in other components as well.

rodrigopavezi: @andregoncalves awesome. cheers

PR #646, opened 2019-02-11
feat(send): update history feature and fix style warnings

eduardosmoura:


src/renderer/transaction/send/index.jsx
```diff
@@ -179,6 +179,22 @@ const getAbrDateFromTimestamp = timestamp => {
 	return { month, day };
 };
 
+const filterTransactionByToken = (transaction, token) => {
+	let valid = false;
+	switch (token) {
+		case 'KEY':
+			valid = transaction.tokenSymbol === 'KI';
```
rodrigopavezi: @eduardosmoura it should be KEY here as it is the production token. I think another case should be added for KI as it is the test token. Let me know if you need help with that. cheers

eduardosmoura: I will fix it @rodrigopavezi, thanks.

rodrigopavezi: @eduardosmoura awesome fix. cheers

PR #658, opened 2019-02-14
Modals

designhorf:
* Close modal - smaller window size, add logo and centre aligned texts
* All Modals bottom padding


designhorf:
I don't think so. As I see only the Close Modal has specific styles (small buttons, narrow modal width).

PR #662, opened 2019-02-14
Feature/605 transaction history

eduardosmoura:


src/renderer/transaction/transactions-history/index.jsx
```diff
@@ -28,94 +30,201 @@ import {
...
+										</Typography>
 									</TableCell>
 									<TableCell align="right">
-										<Button size="small" disableRipple>
-											<ViewIcon />
-											View
-										</Button>
+										<IconButton className={classes.rightSpace}>
+											<CopyToClipboard
+												text={`https://ropsten.etherscan.io/tx/${
```
rodrigopavezi: @eduardosmoura this should be `https://etherscan.io/tx/` for production and `ropsten` for dev mode. You can check which is currently being used by the wallet at run time through `config.chainId` from `common`. Cheers

src/renderer/transaction/transactions-history/index.jsx
```diff
@@ -28,94 +30,201 @@ import {
...
+														Copy
+													</Typography>
+												</Grid>
+											</CopyToClipboard>
+										</IconButton>
+										<IconButton
+											onClick={e => {
+												window.openExternal(
+													e,
+													`https://ropsten.etherscan.io/tx/${
```
rodrigopavezi: @eduardosmoura the same as above. cheers

PR #663, opened 2019-02-15
Feat/transaction with trezor

rodrigopavezi:


src/renderer/wallet/unlock/trezor/enter-passphrase.jsx
```diff
@@ -0,0 +1,226 @@
...
+			this.setState({ error: this.props.error });
+		}
+	}
+
+	handleVisibility = event => {
+		if (this.state.inputType === 'password') {
+			this.setState({
+				...this.state,
+				inputType: 'text',
+				visibilityComponent: <VisibilityOnIcon />
```
sk91: never seen it component added to state previously, interesting idea

PR #673, opened 2019-02-18
Incorporation

andregoncalves:


andregoncalves:
Let's merge this into dev for demo purposes.
I stubbed out two steps of the process (payment because of the high values, and submit kyc)
There is still work to do:
- KYC requirements window needs UI work and the ability to add/edit attributes/documents
- Some UI content is not yet linked correctly to airtable.
- KYC Template is hard coded
- Submit doesn't seem to be working, the KYC application process is not created in the demo instance

andregoncalves:
@rodrigopavezi don't merge yet.

PR #687, opened 2019-02-19
Feature/607 incorporation payment confirmation

eduardosmoura:


eduardosmoura:
> Looks good, @eduardosmoura, i have two questions :
> 
> * There is still a wip label, do you want to do anything yet before I merge?
> * the new selfkey-ui commit hash is necessary? Is it a commit from selfkey-ui master branch and later than the existing on dev?

Question 1) No, it's all done.
Question 2) Ferenc did some tweaks on selfkey-ui required for this component. Probably it will not work properly if we don't use the correct hash, right?

andregoncalves:
Ok, cool. 
@designhorf please do make sure those changes get merged into the master of selfkey-ui (if not already).

PR #727, opened 2019-02-25
fix(new-address): disable submit button when needed

sk91:
+ disable unlock button when no file selected
+ disable unlock button when password empty
+ display file path to indicate file is selected
+ transform error message 'Key derivation failed - possibly wrong password' -> 'Wrong password. Please try again'


sk91:
screen shot 2019-02-25 at 18 21 28

PR #728, opened 2019-02-25
Fix/628 unlock existing wallet fixes

sk91:
+ hide existing address tab when no wallets
+ disable unlock button if wallet was not chosen
+ disable unlock button if password is empty
+ fix incorrect password message
+ use dll plugin to optimise compile time


andregoncalves:
@sk91 merge at at your own discretion!

PR #757, opened 2019-02-27
feat(custom-tokens): update entire flow and add new modals

eduardosmoura:


eduardosmoura:
@rodrigopavezi just to let you know I talked with @designhorf, and we will need to add a transparent background feature on the common/Modal component.

PR #768, opened 2019-02-28
Fix/759 hardware wallets

rodrigopavezi:


rodrigopavezi:
@andregoncalves here

```
async establish() {
		if (this.isActive()) return this.ctx.token;
		let challenge = await RelyingPartyRest.getChallenge(this.ctx);
		let challengeToken = RelyingPartyToken.fromString(challenge.jwt);
		let signature = await this.identity.genSignatureForMessage(
			challengeToken.data.challenge || challengeToken.data.nonce
		);
		let challengeReply = await RelyingPartyRest.postChallengeReply(
			this.ctx,
			challenge.jwt,
			signature
		);
		let token = RelyingPartyToken.fromString(challengeReply.jwt);
		this.ctx.token = token;
		return token;
	}
```

User needs to confirm nonce signing on the device

andregoncalves:
so, that await call blocks until the user confirms?
Does it show anything on the screen?

rodrigopavezi:
@andregoncalves not yet on the screen. We need to add a try catch here https://github.com/SelfKeyFoundation/Identity-Wallet/blob/71da4380574314ed10c54a83a2638025ee74e185/src/common/kyc/index.js#L216

And route to the right screens

PR #769, opened 2019-02-28
feat: allow unauthenticated relying party kyc chain init

andregoncalves:
We can now load the relying party without forcing authentication.
`await this.props.dispatch(kycOperations.loadRelyingParty('incorporations', false));`
second parameter to false indicates it's unauth session. 
This is crying to be refactored, I will open non-priority issue for later.
cc @rodrigopavezi @sk91


src/main/platform/relying-party.js
```diff
@@ -153,7 +153,7 @@ export class RelyingPartyRest {
 		return request.get({
 			url,
 			headers: {
-				Authorization: this.getAuthorizationHeader(ctx.token.toString()),
+				// Authorization: this.getAuthorizationHeader(ctx.token.toString()),
```
rodrigopavezi: @andregoncalves can you remove instead of commenting out?

src/renderer/marketplace/incorporation/detail/index.jsx
```diff
@@ -329,6 +329,8 @@ class IncorporationsDetailView extends Component {
 		const { selectedTab } = this.state;
 		const { translation, tax } = program;
 
+		console.log(this.props.rp);
```
rodrigopavezi: @andregoncalves remove the logs from here

src/common/kyc/index.js
```diff
@@ -211,39 +211,76 @@ const loadRelyingPartyOperation = rpName => async (dispatch, getState) => {
...
+						name: rpName,
+						lastUpdated: ts
+					},
+					error
+				)
+			);
+		}
+	} else {
+		try {
+			const session = mpService.createRelyingPartySession(config);
```
rodrigopavezi: @andregoncalves I think a better way is to move this to a helper function like await getSession(config, authenticate). Cheers

andregoncalves: Yep, good idea.

src/common/kyc/index.js
```diff
@@ -212,8 +223,8 @@ const loadRelyingPartyOperation = rpName => async (dispatch, getState) => {
 	const config = rp.relying_party_config;
 
 	try {
-		const session = mpService.createRelyingPartySession(config);
-		await session.establish();
+		const session = getSession(config, authenticate);
```
rodrigopavezi: @andregoncalves you might need to add an `await` here. cheers

andregoncalves: 👍

PR #823, opened 2019-03-05
Fix/harsource

rodrigopavezi:


package.json
```diff
@@ -235,8 +266,23 @@
...
+				"socket.io-client",
+				"striptags",
+				"trezor-wallet-provider",
+				"web3-provider-engine",
+				"prop-types",
+				"serialize-error",
+				"moment",
+				"bignumber.js",
+				"connected-react-router",
+				"@material-ui/core",
```
sk91: I think you should filter from this list staff that does not go to main thread

PR #931, opened 2019-03-12
fix(exhanges): separate exchanges by env

sk91:


sk91:
@altninja ^^ :) sorry

PR #939, opened 2019-03-12
fix(logs): log errors correctly

sk91:


src/common/kyc/index.js
```diff
@@ -235,7 +243,7 @@ const getSession = async (config, authenticate, dispatch, hardwareWalletType) =>
 			}
 			await session.establish();
 		} catch (error) {
-			console.log(error);
+			log.error('getSession trezor %s', error);
```
rodrigopavezi: @sk91 it could be ledger too. I suggest changing to hardware wallet instead of trezor. cheers

sk91: fixed

PR #946, opened 2019-03-13
fix(custom-token): update loading control flag

eduardosmoura:


eduardosmoura:
Yeah, @rodrigopavezi, I just did that to record it to Ralph because the loading is so quick, but then I had to do something else, and I forgot to remove it. Sorry about that, my bad.

PR #956, opened 2019-03-13
feat(no-internet): show popup if no network

sk91:


src/main/application/network-service.js
```diff
@@ -0,0 +1,28 @@
+import isOnline from 'is-online';
+import { appOperations } from 'common/app';
+
+const NETWORK_CHECK_INTERVAL = 5000;
+
+export class NetworkService {
+	constructor({ store }) {
```
rodrigopavezi: @sk91 so does is mean we can get other service instances on the constructor of a service with awilix?

PR #967, opened 2019-03-14
fix(trezor): add logging and trezor dependency

rodrigopavezi:

PR #971, opened 2019-03-14
fix(custom-token): add token error message

eduardosmoura:


src/common/tokens/operations.js
```diff
@@ -11,13 +11,26 @@ const loadTokens = () => async (dispatch, getState) => {
 
 const addToken = contractAddress => async dispatch => {
 	const tokenService = getGlobalContext().tokenService;
-	const tokenInfo = await tokenService.getTokenInfo(contractAddress);
-	await tokenService.addToken(tokenInfo);
+	try {
+		const tokenInfo = await tokenService.getTokenInfo(contractAddress);
+		await tokenService.addToken(tokenInfo);
+	} catch (error) {
+		await dispatch(
```
rodrigopavezi: @eduardosmoura could you add logging like I did in this PR https://github.com/SelfKeyFoundation/Identity-Wallet/pull/967?

src/common/tokens/operations.js
```diff
@@ -11,13 +11,26 @@ const loadTokens = () => async (dispatch, getState) => {
...
+			actions.setTokenError(
+				`Sorry, an error occurred while adding the current token: ${error}`
```
rodrigopavezi: @eduardosmoura Can you remove the `${error}` from here. We don't want to show an ugly error message to user. It will be logged to sentry so we can debug it later. Cheers

src/common/tokens/operations.js
eduardosmoura: Yes sure @rodrigopavezi

src/common/tokens/operations.js
eduardosmoura: Yes sure @rodrigopavezi

Issue #983, opened 2019-03-18
Incorporation data issue


rhdelima:
![image.png](https://images.zenhubusercontent.com/5b9ab85a7dfd064d6bed8869/b7f359b5-7589-4b94-8029-743eb956fd49)

![image.png](https://images.zenhubusercontent.com/5b9ab85a7dfd064d6bed8869/07f76e8a-9662-451c-a3ad-4e86bbe050ef)

Change font color for links and link should open on a browser
![image](https://user-images.githubusercontent.com/10294014/54553511-4e39d300-49ed-11e9-93e3-be42e600c0c4.png)

andregoncalves:
1. is actually correct. Estonian characters.
2. Should be fixed with latest airtable updated.
3. WIP, requires code change.

PR #990, opened 2019-03-19
fix: link colors and open external in dynamic airtable HTML #983

andregoncalves:
Fixes issues with anchor tags in the dynamically loaded html from airtable.
- Color of links
- Links open in external browser window


src/renderer/marketplace/incorporation/detail/index.jsx
```diff
@@ -155,6 +155,9 @@ const styles = theme => ({
 		'& ul li': {
 			lineHeight: '1.4em',
 			marginBottom: '0.5em'
+		},
+		'& a': {
+			color: '#93B0C1'
```
sk91: @andregoncalves maybe better: 
`color: theme.palette.secondary.main`
or `import {typography} from 'selfkey-ui"`
`color: typography`

https://github.com/SelfKeyFoundation/selfkey-ui/blob/master/src/colors.tsx#L3
https://github.com/SelfKeyFoundation/selfkey-ui/blob/master/src/theme/selfkey-dark-theme.tsx#L48

rodrigopavezi:
@andregoncalves would it make your life easier when our electron app will open all target="_blank" links externally. Because I am adding the bellow for that.

```
mainWindow.webContents.on('new-window', function(e, url) {
  e.preventDefault();
  require('electron').shell.openExternal(url);
});
```
Cheers

andregoncalves:
@rodrigopavezi that definitely makes it easier. Is this is going into the release right now? Should I remove the click capture and handling?

rodrigopavezi:
@andregoncalves oh, do you need it for this release? I thought otherwise because you are merging to dev branch. we can add it to this release for sure and you can remove the handling here. Cheers

src/renderer/marketplace/incorporation/detail/index.jsx
andregoncalves: Got it.

PR #999, opened 2019-03-20
fix(not-closing-modal): remove the cleanup  of rp session when timeout

rodrigopavezi:


rodrigopavezi:
@andregoncalves it seems to be added to all the error on transaction which I think is not correctly because if you go back from the timeout screen the session should still be open. In other case if there is a transaction error. That will send the user to dashboard so I agree that we need to clear session. Cheers

PR #1022, opened 2019-03-26
fix(e2e-test): get e2e back to work

eduardosmoura:


.circleci/config.yml
```diff
@@ -260,9 +260,9 @@ workflows:
...
                       - unit-test
-            #            - 'e2e-mac':
-            #                  requires:
-            #                      - build-mac
+            - 'e2e-mac':
+                  requires:
+                      - build-mac
             - 'stage-mac':
                   requires:
                       - build-mac
```
rodrigopavezi: @eduardosmoura can you change to only stage-mac if e2e-mac passes. You need just to change the `build-mac` to `e2e-mac`

Also can you add a job to run e2e on linux and also require it to stage-linux?

Cheers

eduardosmoura: Will do @rodrigopavezi

PR #1107, opened 2019-04-24
Feat/1097 signup

rodrigopavezi:


rodrigopavezi:
@andregoncalves I have changed the vendor values for incorporation to be passed as params to the current application. Please review. cheers

Issue #1133, opened 2019-04-29
Fix inconsistent buttons


oanamangiurea:
We have a few buttons in the app that are inconsistent in terms of size and font-size. They should be fixed, make sure they use the already defined ones in the Selfkey UI, and follow the design.

What I found so far:

- [x] For all buttons -> Add letter-spacing: 0.6px as it was in original design (especially because on how poor Lato renders on retina displays) 
- [x] For all buttons -> change min-width to 120px instead of 200px
- [x] Transaction modal buttons https://gyazo.com/e2af75ddd8fbbb743c4dd08acb48749c - font size is incorrect -> should be 16px
- [x] Pending Key Deposit button -> https://gyazo.com/6138de1d09993f58e1c151a10ab438ad -> please style the icon as in the design https://projects.invisionapp.com/d/main#/console/15222689/347983074/preview and space it from the text 
- [x] Buttons text are not vertically aligned -> https://gyazo.com/428457e05c52c535159a1260f469f2f9 Fix: add line height on the that has the text inside the buttons. Please check all buttons for this issue

CortecKYC:
@designhorf Can you please estimate this issue?

designhorf:
@CortecKYC sure, I'll add an estimation soon

PR #1148, opened 2019-05-01
Fix/1133 incosistent buttons

designhorf:
Changes: 
https://github.com/SelfKeyFoundation/Identity-Wallet/issues/1133


designhorf:
@sk91 updated

PR #1173, opened 2019-05-10
feat(create-did): for hardware wallets

rodrigopavezi:


rodrigopavezi:
@eduardosmoura there is an impact and I updated the test to account for that. cheers

PR #1177, opened 2019-05-13
Feature/1063 associate did

eduardosmoura:


src/renderer/selfkey-id/main/components/associate-did.jsx
```diff
@@ -0,0 +1,254 @@
...
+													Please wait. Checking the blockchain for DID{' '}
+													information.
+												</span>
+											</React.Fragment>
+										)}
+									</Typography>
+									<Input
+										placeholder="did:selfkey:"
+										name="did"
+										value={did}
```
rodrigopavezi: @eduardosmoura could you also set the error attribute of the Input field so it will highlight the field. Cheers

eduardosmoura: @rodrigopavezi In fact, the "className={didInputClass}" attribute will do that highlight. 

const didInputClass = `${classes.input} ${hasAssociateError && !searching ? classes.errorColor:''}`;

PR #1185, opened 2019-05-15
feat(remove-staking): from react front end

rodrigopavezi:


src/renderer/marketplace/services/service-details.jsx
```diff
@@ -269,8 +269,6 @@ class MarketplaceServiceDetailsComponent extends Component {
 		const application = this.getLastApplication();
 		// FIXME: Troubleshooting and overriding
```
sk91: please remove comments as well

Issue #1102, opened 2019-04-23
Transaction in History is only displayed when it is completed. So Sending/Receiving status won't be shown.


rodrigopavezi:

rhdelima:
Looks good. See sample screenshots.

![image](https://user-images.githubusercontent.com/10294014/59668497-325fec00-91eb-11e9-8cd9-4e66345e81c1.png)

![image](https://user-images.githubusercontent.com/10294014/59668699-836fe000-91eb-11e9-9e0c-ed3352122509.png)

![image](https://user-images.githubusercontent.com/10294014/59671125-ae5c3300-91ef-11e9-8b6a-26893c6fe749.png)

PR #1301, opened 2019-06-11
fix: update transaction history to display sending status

maycon-mello:
fix for #1102

I've updated the transaction history service. The old implementation was removing all pending transactions from the list, and also was not receiving updates from the blockchain. The user would have to hard reload the list to see new transactions.

The new implementation should:
* Keep pending transactions in the db
* Check all the pending transactions in the blockchain to receive updates
* Remove failed transactions from the db.
* I've also fixed a bug related to uppercase "from" and "to" attributes being saved into the db. It was breaking the search results.

### DIFF
--- a/src/main/blockchain/tx-history-service.js
+++ b/src/main/blockchain/tx-history-service.js
@@ -316,12 +316,7 @@ export class TxHistoryService {
 		}
 	}
 	async removeNotMinedPendingTxs(address) {
-		let nonce = await this.web3Service.waitForTicket({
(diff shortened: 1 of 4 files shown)


maycon-mello:
> @maycon-mello why are you removing failed transactions? I think we should show them as well as it is a transaction history. Cheers

@rodrigopavezi sure, the old implementation used to remove it so I decided to keep it. I've updated the implementation to not remove failed transactions anymore. Please review.

Regards,

PR #1304, opened 2019-06-11
fix(manage-application): for hardware wallets

rodrigopavezi:


src/renderer/selfkey-id/main/components/selfkey-id-applications.jsx
```diff
@@ -308,12 +308,6 @@ class SelfkeyIdApplicationsComponent extends Component {
 		if (!this.props.incorporations || !this.props.incorporations.length) {
```
andregoncalves: This will need refactoring later on.
In a perfect world, an it should be agnostic of any specific rp.

Issue #1286, opened 2019-06-04
Handle ETIMEDOUT error properly


rodrigopavezi:
![Screenshot 2019-06-04 at 11.16.25.png](https://images.zenhubusercontent.com/5b7bd3b852bce06c3c5c53c0/780d62db-3bc3-42b4-bed2-43765f15500f)
![Screenshot 2019-06-04 at 11.17.42.png](https://images.zenhubusercontent.com/5b7bd3b852bce06c3c5c53c0/04923812-ff50-4972-a749-e8ad2d58000b)

PR #1315, opened 2019-06-13
fix: correct loading handling for marketplaces

andregoncalves:
#1286


andregoncalves:
This screen is now shown if data cannot be loaded.
Screen Shot 2019-06-13 at 11 59 10

PR #1332, opened 2019-06-19
fix(custom-token): abi symbol output

rodrigopavezi:


rodrigopavezi:
@andregoncalves good catch. Added a link to the stackoverflow issue

PR #1355, opened 2019-06-25
Feat/merge back

rodrigopavezi:


src/renderer/selfkey-id/main/components/selfkey-id-applications.jsx
```diff
@@ -261,9 +271,254 @@ const getRpInfo = (rpName, field) => {
 const MARKETPLACE_ROOT_PATH = '/main/marketplace-categories';
 
 class SelfkeyIdApplicationsComponent extends Component {
```
sk91: @rodrigopavezi are you sure this is right? I wouldnt expect ui to be only on release branch maybe it is overriding some changes we did in dev branch?

src/renderer/selfkey-id/main/components/selfkey-id.jsx
```diff
@@ -1,302 +1,38 @@
 import React, { Component } from 'react';
-import CircularProgress from '@material-ui/core/CircularProgress';
-import { Grid, Typography, Tabs, Tab, Button, withStyles } from '@material-ui/core';
-import { connect } from 'react-redux';
```
sk91: Same here about ui

src/renderer/selfkey-id/main/components/selfkey-id-applications.jsx
rodrigopavezi: @sk91 that is correct. It was only changed on the release branch and there was no merging issue when I merge dev into this branch.

src/renderer/selfkey-id/main/components/selfkey-id.jsx
rodrigopavezi: Same here too

PR #1391, opened 2019-07-03
fix(ui): remove back button from confirm screen

sk91:


andregoncalves:
@sk91 is this fix in dev?

PR #1395, opened 2019-07-04
feat: wire-up to payment container

andregoncalves:


src/common/marketplaces/orders/index.js
```diff
@@ -540,11 +541,7 @@ const ordersSelectors = {
 	getOrderPriceUsd: (state, id) => {
 		const order = ordersSelectors.getOrder(state, id);
 		let fiat = fiatCurrencySelectors.getFiatCurrency(state);
-		let fiatRate = pricesSelectors.getRate(
-			state,
-			config.constants.primaryToken,
-			fiat.fiatCurrency
-		);
+		let fiatRate = pricesSelectors.getRate(state, 'KEY', fiat.fiatCurrency);
```
sk91: use `config.constants.primaryToken` instead of 'KEY' to use KI on dev

src/renderer/marketplace/bank-accounts/checkout/payment-container.jsx
```diff
@@ -61,6 +61,7 @@ const mapStateToProps = (state, props) => {
 		accountType: bankAccountsSelectors.getTypeByAccountCode(state, accountCode),
 		banks: bankAccountsSelectors.getDetailsByAccountCode(state, accountCode),
 		publicKey: getWallet(state).publicKey,
+		keyRate: pricesSelectors.getRate(state, 'KEY', 'USD'),
```
sk91: same here with primary token

PR #1400, opened 2019-07-09
fix(manage-applications): update rpName and messages

eduardosmoura:


src/renderer/selfkey-id/main/components/selfkey-id-applications.jsx
```diff
@@ -268,6 +267,10 @@ const getRpInfo = (rpName, field) => {
 	return config.relyingPartyInfo[rpName][field];
 };
 
+const getRpName = title => {
+	return title.startsWith('Bank Account') ? 'Bank Accounts' : 'Incorporations';
```
andregoncalves: It's probably better to do `title.toLowerCase().startsWith('bank account')` to avoid a case mismatch.

eduardosmoura: Good catch. It's done now @andregoncalves.

eduardosmoura:
Comment added as well @andregoncalves. Thank you.

PR #1408, opened 2019-07-11
refactor: payment amount

andregoncalves:
Refactor contract amount to own selector.
Fix problem when allowance is too large (53bit issue)
Fiat rate  needs to be in 'KEY'

### DIFF
--- a/src/common/marketplaces/orders/index.js
+++ b/src/common/marketplaces/orders/index.js
@@ -217,19 +217,14 @@ const checkOrderAllowanceOperation = orderId => async (dispatch, getState) => {
 	);
 	let update = null;
 
-	try {
-		allowance = allowance.toNumber();
-	} catch (error) {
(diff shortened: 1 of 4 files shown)


src/common/marketplaces/orders/index.js
```diff
@@ -213,18 +209,27 @@ const checkOrderAllowanceOperation = orderId => async (dispatch, getState) => {
...
+	const amount = ordersSelectors.getContractFormattedAmount(getState(), orderId);
+
+	let allowance = await selfkeyService.getAllowance(
 		wallet.publicKey,
 		config.paymentSplitterAddress
 	);
 	let update = null;
 
-	if (allowance.toNumber() < order.amount && order.status === orderStatus.ALLOWANCE_COMPLETE) {
+	try {
```
sk91: That is a miscalculation on my part, we are better off using BN here

PR #1430, opened 2019-07-15
fix(no-application): change to close the wallet

eduardosmoura:

### DIFF
--- a/src/renderer/no-connection/index.jsx
+++ b/src/renderer/no-connection/index.jsx
@@ -28,8 +28,10 @@ const styles = theme => ({
 });
 
 export const NoConnection = withStyles(styles)(props => {
-	const { classes, onBackClick = false, dispatch } = props;
-	const backToDashboard = () => dispatch(push(`/main/dashboard`));
+	const { classes, onBackClick = false } = props;
+	const handleClose = () => {
+		window.quit();
+	};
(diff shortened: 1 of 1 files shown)


eduardosmoura:
@sk91 if you take a look at the design (https://projects.invisionapp.com/d/main#/console/15222689/319048193/preview), that action (close) makes sense to be done. Maybe we should change the text description as well if we decide to wait for the connection to be restored. Do we need Oana's help on this one?

sk91:
@eduardosmoura @designhorf @oanamangiurea 

Here is the original version of the screen
https://github.com/SelfKeyFoundation/Identity-Wallet/commit/f00dc9c7af23cd6a0ad6c0ba38b0e8826d406b70

the text is correct and no button.

I remember us discussing this version with @oanamangiurea when we developed the screen

What probably happened is that we made a decision to do it like this, and not changed the design in invision

I would prefer to revert to this version if possible unless @oanamangiurea and you guys disagree

designhorf:
@max @eduardosmoura @oanamangiurea 
`Please check your connection and reopen the application.` This is the second sentence of the popup. It means (for me), I have to close the application and if the connection is restored, open it again. In this case I think the button is justifiable but without quit from application trigger.

PR #1470, opened 2019-07-29
feat(inventory-fetcher): add flagtheory incorporations inventory fetched

sk91:


sk91:
Added bank accounts and fixed bugs

All data in redux now @andregoncalves

Issue #1321, opened 2019-06-17
Incorporations Marketplace Refactor to use same architecture as Bank Accounts


andregoncalves:
Refactor components using the same architecture as Bank Accounts. Create Storybook entries and rename components and folders. Extract common components to /marketplace/common.

CortecKYC:
@maycon-mello Have you finished this card? If so, please move it to the next pipeline

maycon-mello:
> @maycon-mello Have you finished this card? If so, please move it to the next pipeline

This is almost done. I'm applying some banking/accounts updates the team did on past days. I will submit the PR for review by the end of the day.

andregoncalves:
@rhdelima we refactored the incorporations application flow with improvements we did for Bank Accounts. This requires a round QA to make sure we didn't break anything.

rhdelima:
@andregoncalves A blank screen is displayed after submitting requirements.

![image](https://user-images.githubusercontent.com/10294014/62832571-4eb85c00-bc63-11e9-8a27-a8258a057c4f.png)

![image](https://user-images.githubusercontent.com/10294014/62832587-8d4e1680-bc63-11e9-93d7-8ef50e425f8f.png)

andregoncalves:
I was able to reproduce it, it's related to the scheduler work that's not going into this release.
I'm working on a fix.

PR #1471, opened 2019-07-30
Refactor/incorporations views

andregoncalves:
Related to #1321 
Refactors incorporations flow to be similar and use the same common components as Bank Accounts Marketplace.


andregoncalves:
@sk91 sorry for the big pull request, I wanted to get this done before integrating the scheduler.

PR #1485, opened 2019-08-12
fix: marketplace orders integration with scheduler feature flag

andregoncalves:
Disabling the scheduler via feature flag was breaking the marketplace orders loading.
This fixes it.


src/renderer/marketplace/bank-accounts/index.jsx
```diff
@@ -9,7 +12,14 @@ import { BankAccountsPaymentCompleteContainer } from './checkout/payment-complet
 import { BankAccountsSelectBankContainer } from './select-bank/select-bank-container';
 import { BankAccountsProcessStartedContainer } from './process-started/process-started-container';
 
-class MarketplaceBankAccountsPage extends Component {
+class MarketplaceBankAccountsComponent extends Component {
+	async componentDidMount() {
+		if (featureIsEnabled('scheduler')) {
```
sk91: New marketplace data structure should not depend on scheduler feature.

Scheduler's only function is to perform jobs, it is used to populate this data, yes, but the structure itself should remain the same whether the scheduler is enabled or if it's disabled and you load data manually

andregoncalves: This is not for the new marketplace data refactoring, this is for the local orders model.
If the scheduler is not active, the orders are not loaded, see:
```
const loadMarketplaceOperation = () => async (dispatch, getState) => {
	await Promise.all([
		dispatch(countriesOperations.loadCountriesOperation()),
		dispatch(taxTreatiesOperations.loadTaxTreatiesOperation()),
		dispatch(inventoryOperations.loadInventoryOperation()),
		dispatch(vendorOperations.loadVendorsOperation()),
		dispatch(ordersOperations.ordersLoadOperation())
	]);
};
```
This probably should be a separate feature flag, but for now if we disable scheduler the orders are not loaded to the store.

sk91: Let's sync on it after standup to be on the same page

PR #1538, opened 2019-08-19
fix(address-book): remove duplicates

eduardosmoura:


sk91:
@eduardosmoura is this ready for review?

src/renderer/address-book/main/index.jsx
```diff
@@ -69,7 +75,7 @@ class AddressBookContainer extends Component {
 
 	render() {
 		const { classes } = this.props;
-		const { addresses } = this.state;
+		const addresses = removeDuplicates(this.state.addresses, 'address');
```
sk91: seems very complicated, maybe better like this using lodash:
`const addresses = _.uniqueBy(this.state.addresses, 'address');`

PR #1577, opened 2019-09-05
feat: performance improvements when syncing large amounts of data

andregoncalves:
Tackles issue #1567.
Decided to not include an external dependency.
Tests indicate performance issue is gone.


src/main/marketplace/tax-treaties/tax-treaties-service.js
```diff
@@ -13,6 +13,17 @@ export const TAX_TREATIES_API_ENDPOINT = `${config.airtableBaseUrl}TaxTreaties${
 	isDevMode() ? 'Dev' : ''
 }`;
 
+// Helper function to evenly distribute in time expensive background processing.
+// This avoids blocking the event loop, and should be used when we are expecting
+// to parse and process a very large number of (non time-critical) items
+const executeAsync = (idx, fx) =>
+	new Promise((resolve, reject) =>
+		setTimeout(() => {
```
sk91: better use setImmediate if it's available

src/main/marketplace/tax-treaties/tax-treaties-service.js
```diff
@@ -13,6 +13,17 @@ export const TAX_TREATIES_API_ENDPOINT = `${config.airtableBaseUrl}TaxTreaties${
...
+			const item = fx();
+			resolve(item);
+		}, idx * 10)
```
sk91: index * 10 ?

you don't need to worry about finish time since you use `Promise.all` anyway, also all timers are executed via queue so your order will be preserved anyway if you set it to the same timeout. Moreover, you don't need to worry about timeout if you use setImmediate

andregoncalves:
After testing it more carefully, the performance issues were caused by several factors:
- Processing large amount of data in the same event loop
- Large sqlite database changes also blocking the event loop
- Sqlite3 database pool exhausted because scheduler sync actions were being executed almost all at roughly the same time.

andregoncalves:
Closing this pull request please see https://github.com/SelfKeyFoundation/Identity-Wallet/pull/1580

PR #1580, opened 2019-09-08
Fix/async tax treaties loading

andregoncalves:
After testing it more carefully, the performance issues were caused by several factors:

Processing large amount of data in the same event loop
Large sqlite database changes also blocking the event loop
Sqlite3 database pool exhausted because scheduler sync actions were being executed almost all at roughly the same time.


src/main/marketplace/tax-treaties/tax-treaties-service.js
```diff
@@ -58,26 +62,40 @@ export class TaxTreatiesService {
...
+		const chunks = arrayChunks(upsert, 100);
+		for (const c of chunks) {
+			TaxTreaties.bulkUpsert(c);
+			await setImmediatePromise();
+		}
 	}
-	deleteMany(ids) {
-		return TaxTreaties.deleteMany(ids);
+	async deleteMany(ids) {
+		TaxTreaties.deleteMany(ids);
```
sk91: should have `return` or `await` near `TaxTreaties.deleteMany(ids)`, otherwise it will not wait for promise

Also, it might be beneficial to use array chunks here as well, becouse sometimes it fails on too many ids

src/main/marketplace/tax-treaties/tax-treaties-service.js
```diff
@@ -58,26 +62,40 @@ export class TaxTreatiesService {
...
-
 	loadTaxTreaties() {
 		return TaxTreaties.findAll();
 	}
-	upsert(upsert) {
-		return TaxTreaties.bulkUpsert(upsert);
+	async upsert(upsert) {
+		const chunks = arrayChunks(upsert, 100);
+		for (const c of chunks) {
+			TaxTreaties.bulkUpsert(c);
```
sk91: should be `await` here, the promise should not be ignored

sk91: Also, maybe all of them need to be wrapped in a transaction? So that if upsert of one chunk fails, all of the operation will be failed ?

sk91: Another thing, it seems to me like a model level operation, instead of service level. It is atomic to `bulkUpsert` operation in my opinion

andregoncalves: Agree 💯 , the service should be agnostic to the upsert model implementation details (and a yes a transaction was clearly missing here).

PR #1588, opened 2019-09-13
feat(attr-types): use only individual attributes for individual profiles

sk91:


sk91:
for qa:

Open individual profile, and see that the user can fill only individual identity attributes

A list is here: https://platform.selfkey.org/repository.json (has 'individual' in entityType attribute)

Issue #1552, opened 2019-08-23
CORP1: UI add entity switcher to menu


sk91:

eduardosmoura:
@oanamangiurea do we have screens (design) for this ticket? 

@sk91 what's the scope for this ticket? Is it just the "Wallet Menu"? Or it should go all the way down to "New Corp Wizard" and "Corp Profile - Overview"? 

https://confluence.kyc-chain.com/pages/viewpage.action?spaceKey=SEL&title=CORP1%3A+Create+New+Company+Profile

oanamangiurea:
@eduardosmoura yes we do, if think ticket refers to the following: 

https://projects.invisionapp.com/d/main#/console/15222689/375497853/preview - see top right name and dropdown

And dropdown itself https://projects.invisionapp.com/d/main#/console/15222689/380279179/preview -> where the button new corporate profile will go to https://projects.invisionapp.com/d/main#/console/15222689/316560291/preview

eduardosmoura:
Thank you so much @oanamangiurea

oanamangiurea:
@sk91 I meant click on the name that is in the top bar, not in the dropdown menu. For sure the dropdown menu should change accounts. Still UI needs to look like in design

sk91:
Oh, sorry, got confused here for a sec :)

CortecKYC:
If Ferenc still has to work on this ticket, then it's not QA ready. Can you guys confirm?
@oanamangiurea @eduardosmoura

rhdelima:
@designhorf Failing this one due to the following issues:
1. User name is not updated after adding SelfKey ID. Still showing `New Individual`
2. `New Corporate Profile` is not working
![image](https://user-images.githubusercontent.com/10294014/65314680-5efc1900-dbc9-11e9-9cb1-2550e07de843.png)

build: dev_2019.09.19.12

oanamangiurea:
@designhorf  also can you please to the following updates:

https://gyazo.com/7980a9b39d7730921090dff3b43d31b0

1. Button small- let's decrease the font size from 12px to 10px, with character spacing of 0.4. I've updated the style guide as well 
2. Button should have same with as separator line 
3. For the dropdown - add more spacing between the User name and The profile type . Also the font sizing is too big, one should be 14px, the other one explanatory with 13px
[...]

PR #1607, opened 2019-09-24
Fix/1552 entity switcher to menu

designhorf:
Changes:

- Fixed these issues https://github.com/SelfKeyFoundation/Identity-Wallet/issues/1552#issuecomment-533522104


andregoncalves:
@designhorf can you open a separate issue for the username?

designhorf:
@andregoncalves sure

designhorf:
Created separate issue for username

src/renderer/wallet/main/toolbar.jsx
```diff
@@ -167,10 +183,10 @@ const ProfileList = withStyles(profileStyle)(
 									)}
 								</Grid>
 								<Grid item sm={8} className={classes.profileName}>
-									<Typography variant="h6">
+									<Typography variant="subtitle1" style={{ marginBottom: '5px' }}>
```
andregoncalves: It's better to try and avoid inline styles as much as possible.
I would suggest moving this to the profileName css rules.

PR #1647, opened 2019-10-01
Feat/1634 hd applications

sk91:
- fix kyc-applications
- fix manage applications for hw wallets
- fix switch web3 provider between default and hw wallets


andregoncalves:
Applications › Operations › loadApplicationsOperation test seems to be failing:
src/common/kyc/kyc-duck.spec.js line 214

PR #1697, opened 2019-10-07
Fix/1628 drop down

designhorf:
Changes: 

- Removed selected profile from drop down


src/renderer/wallet/main/toolbar.jsx
```diff
@@ -207,44 +207,50 @@ const ProfileList = withStyles(profileStyle)(
...
-										) : el.profilePicture ? (
-											<HexagonAvatar
-												src={el.profilePicture}
-												className={classes.profileListAvatar}
-												smallAvatar={true}
-											/>
-										) : (
-											<SmallRoundPerson />
-										)}
+							profiles.map((el, index) =>
```
andregoncalves: I think It's more legible and clean filtering the array before executing the map
`profiles.filter(el => el.id !== selectedProfile.id).map(...`

PR #1700, opened 2019-10-08
Order bank offers alphabetically

### LINKED ISSUE #1587: Bank Account - country order
In the newly updated bank accounts country lists, the countries should be ordered alphabetically, on all 3 tabs

https://gyazo.com/128996ea89f7ffa5f68c30778dd89752

andregoncalves:
Fixes #1587


andregoncalves:
> Looks good.
> 
> In terms of performance, it's better to filter first, than sort

Yes, Nice catch. I swapped the call order.

Issue #1573, opened 2019-09-04
Exchanges: Enable affiliate links to Crypto-exchanges and add affiliate disclaimer


CortecKYC:
- List of affiliate links: https://confluence.kyc-chain.com/display/SEL/Exchanges+Development+airtable (only 3 for now)
- [ ] Enable affiliate links to Exchanges in the list above
- [ ] Add affiliate disclaimer (users must be aware that they may click an affiliate link and we may earn commissions for it)

andregoncalves:
@CortecKYC where do we have the affiliate links for the exchanges? 
The URLs in the airtable don't seem to be affiliate links.

sk91:
@andregoncalves  https://confluence.kyc-chain.com/display/SEL/Exchanges+Development+airtable

CortecKYC:
@andregoncalves It was in the ticket description ;)

andregoncalves:
Ah correct, sorry.

andregoncalves:
A new column (affiliateURL) in the exchanges airtable is used for loading the url target for the button. Nothing is displayed if no affliateURL exists.

![Screenshot 2019-10-10 at 20.52.49.png](https://images.zenhubusercontent.com/5be15edea977220c77bd265d/f7dca2b9-f2f0-4081-9675-93fdf66e8f04)

rhdelima:
Looks good. Pass

CortecKYC:
@andregoncalves @rhdelima I just installed 4.1.0 and clicked on an affiliate link. I did not see any disclaimer disclosing that the link was an affiliate and what it means. 
Where is it?

andregoncalves:
I must have missed that... the only disclaimers included are the ones described here:
https://github.com/SelfKeyFoundation/Identity-Wallet/pull/1710
(on the listing and on the detail page).

Where is the text for the affiliate specific disclaimer?

oanamangiurea:
@andregoncalves here is the text and the screen. Placed text under the button https://projects.invisionapp.com/d/main#/console/15222689/389299195/preview

CortecKYC:
New ticket: https://github.com/SelfKeyFoundation/Identity-Wallet/issues/1746

PR #1711, opened 2019-10-10
feat: affiliate link for exchanges

andregoncalves:
Affiliate signup button for exchanges. #1573 
Screenshot 2019-10-10 at 20 52 49


andregoncalves:
A new column - affiliateURL - in the exchanges airtable is used for loading the link.

PR #1757, opened 2019-10-21
Certifiers

sk91:


sk91:
@designhorf please have a pull request for each branch that is work in progress or delete it if it is obsolete

designhorf:
Changes: 

- Selfkey UI updated
- Route for Notarise Marketplace first screen
- Use test data for content

Issue #1679, opened 2019-10-07
CORP3: ui manage members tab


sk91:
- Implement ui for manage members tab

rhdelima:
Failing this one. Please see issues below:
1. Name should be displayed on one line
2. Missing email
3. Share should be in %
4. Table is NOT shown properly. Some parts are hidden
![image](https://user-images.githubusercontent.com/10294014/69826353-b5cb4200-124d-11ea-96b2-a85d808acc65.png)

andregoncalves:
1, 2 and 3 addressed on #1921.
@designhorf is working separately to fix 4.

rhdelima:
`Shares` still not showing % but will log it on separate ticket. Closing this one.

PR #1787, opened 2019-10-29
feat: corporate member ui and storybook

andregoncalves:
#1679 
Screenshot 2019-10-29 at 10 23 17


andregoncalves:
This will need work later on when wiring-up to redux store

Issue #1684, opened 2019-10-07
CORP3: dashboard - display shareholders/members equity graph if there are members with equity


sk91:

rhdelima:
Failing this one. 
1. Graph is incorrect. It shows 100% instead of 35%
2. Should display `First Name Last Name` instead of `Last Name, First Name`

![image](https://user-images.githubusercontent.com/10294014/69849898-e85e5480-12b8-11ea-8cac-b339846766f8.png)

rhdelima:
Failing this one again. As confirmed with Oana, please see issues below:
1. Unassigned shares should have a label `Unassigned` on graph
2. Color for unassigned share should be gray (#999999)
![image](https://user-images.githubusercontent.com/10294014/70304933-00e0e880-183e-11ea-95a8-0780c1f3f273.png)

sk91:
<img width="942" alt="Screenshot 2019-12-09 at 11 47 22" src="https://user-images.githubusercontent.com/1021965/70425416-d3807e80-1a79-11ea-8dbd-6297f0b3fe63.png">

rhdelima:
Looks good. But will log a separate ticket for issue on the tooltip. Should only show shareholder and % of shares (ex. John Doe 30%). Currently, tooltip shows this format `John Doe 30 (30%)`

![image](https://user-images.githubusercontent.com/10294014/70562835-22d4c500-1bc8-11ea-871a-2c5bd289ce8c.png)

PR #1798, opened 2019-10-31
feat: adjust corporate shareholding graph to the new members structure

andregoncalves:
#1684 

Screenshot 2019-10-31 at 09 57 14

### DIFF
--- a/stories/corporate-data.js
+++ b/stories/corporate-data.js
@@ -562,5 +562,36 @@ export const corporateMembers = [
 				position: 'observer'
 			}
 		]
+	},
+	{
+		entity: {
+			type: 'corporate',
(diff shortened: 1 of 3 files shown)


stories/corporate-data.js
```diff
@@ -562,5 +562,36 @@ export const corporateMembers = [
...
+			companyName: 'Selfkey Org',
+			email: 'info@selfkeyorg'
+		},
+		attributes: [
+			{
+				title: 'incorporation',
+				name: 'incorporation',
+				type: {
+					content: {
+						$id: 'http://platform.selfkey.org/schema/attribute/legal-jurisdiction.json',
```
sk91: We should really start tracking the attributes that we are explicitly using somewhere in constants, separate issue though

PR #1823, opened 2019-11-11
memoize all identity selectors

sk91:


sk91:
FOR QA:

Please perform general sanity check for the wallet

PR #1839, opened 2019-11-13
Feat/certifiers messages

designhorf:
Changes:
- Messages UI screen added to Storybook


src/renderer/marketplace/notarization/common/message-container.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+const styles = theme => ({});
+
+class NotarizationMessageContainer extends PureComponent {
+	render() {
+		const { onBackClick, messages } = this.props;
+		return <NotarizationMessageWidget onBackClick={onBackClick} messages={messages} />;
+	}
+}
+
```
andregoncalves: Simple functional components should use functions instead of classes for simplicity, see https://reactjs.org/docs/components-and-props.html#function-and-class-components.

Unless you believe this component will in the future need complex lifecycle functionality and/or local state

sk91: @andregoncalves @designhorf it's better to keep it as class because it's a container and our version of react does not support hooks

A better question would be "Should this be a container?"

stories/notarization.stories.js
```diff
@@ -5,6 +5,7 @@ import NotarizationDetailsPage from '../src/renderer/marketplace/notarization/de
 import RequestNotarizationPage from '../src/renderer/marketplace/notarization/process/request-notarization-page';
 import TOCPopup from '../src/renderer/marketplace/notarization/common/toc-popup';
 import TOCDisagreementPopup from '../src/renderer/marketplace/notarization/common/toc-disagreement-popup';
+import NotarizationMessageContainer from '../src/renderer/marketplace/notarization/common/message-container';
```
sk91: Containers should not be in storybook

PR #1853, opened 2019-11-18
Feat/delete member wire up

andregoncalves:
Delete corporate member operation wire-up


src/renderer/corporate/member/corporate-add-member-container.jsx
```diff
@@ -85,6 +137,37 @@ class CorporateAddMemberContainer extends PureComponent {
...
+
+	validateAttributeDid = did => true;
+
+	filterAcceptablePositions = selectedPositions =>
+		selectedPositions.filter(p =>
+			this.props.availablePositions.map(pos => pos.position).includes(p)
+		);
+
+	validateAttributeParentId = parentId =>
+		this.props.companies.find(c => c.identity.id === parentId);
```
sk91: does it work? I can see an issue here with parentId type, in identity it's integer and in form it might be 'string'

andregoncalves: It works... it's javascript! 
Fixed on fdfa38c

PR #1858, opened 2019-11-18
feat: refactor link to account for non-affiliates

andregoncalves:


src/renderer/marketplace/exchanges/exchanges-details.jsx
```diff
@@ -319,7 +322,7 @@ class ExchangesDetailsComponent extends PureComponent {
 		}
 	};
 
-	renderAffiliateLinkButton = url => (
+	renderLinkButton = ({ url, isAffiliate = false, text = 'SIGN UP' }) => (
```
sk91: In general it is better to do it as a component instead of a render method

andregoncalves: Improved on 8d9286b

Issue #1865, opened 2019-11-19
Member equity validation


andregoncalves:
User CorporateStructureSchema.getEquityForCompanyType

rhdelima:
@andregoncalves Can you provide some instruction on how to test this? Thanks!

andregoncalves:
Some positions require equity, some don't, that needs to be tested.
Also, we need to test that only valid equity values are allowed.

rhdelima:
@andregoncalves Do you have list of roles require equity? Based on my testing, only `Member` role has shares. Also, I can save a member (with role = `Member`) with 0 equity. Shouldn't we require `> 0` equity? Thanks!

sk91:
```
	'director-ltd': false,
	'director-fnd': false,
	shareholder: 'Shares',
	ubo: false,
	observer: false,
	authorizedSignatory: false,
	manager: false,
	member: 'Membership Interest',
	'member-llc': 'Membership Interest',
	grantor: false,
	'beneficiary-tst': false,
	'beneficiary-fnd': false,
	trustee: false,
	protector: false,
	founder: false,
	'general-partner': 'Percentage',
	'limited-partner': 'Percentage'
```

Every position that does not have "false" has equity

rhdelima:
Looks good. Pass

PR #1874, opened 2019-11-25
feat: add equity validation based on position

andregoncalves:
#1865

### DIFF
--- a/src/renderer/corporate/member/member-shares-form.jsx
+++ b/src/renderer/corporate/member/member-shares-form.jsx
@@ -90,6 +90,8 @@ const CorporateMemberSharesFormComponent = withStyles(styles)(props => {
 	const {
 		classes,
 		companies = [],
+		positions = [],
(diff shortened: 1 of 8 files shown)


andregoncalves:
Fixing the tests now

src/renderer/corporate/member/member-shares-form.jsx
```diff
@@ -133,6 +135,7 @@ const CorporateMemberSharesFormComponent = withStyles(styles)(props => {
 							required
 							error={errors.equity}
 							value={equity}
+							disabled={!positions.some(p => positionsWithEquity.includes(p))}
 							onChange={onFieldChange('equity')}
```
sk91: LGTM

andregoncalves: That line is difficult to grasp.
It's probably better to have an helper function for that.

PR #1884, opened 2019-11-26
fix: improved member validation and creation

andregoncalves:
* Fix wrong required attributes
* Move attribute required checking to before identity creation
* Validate empty member position

### DIFF
--- a/src/renderer/corporate/member/corporate-member-container.jsx
+++ b/src/renderer/corporate/member/corporate-member-container.jsx
@@ -165,7 +165,6 @@ class CorporateMemberContainerComponent extends PureComponent {
 			throw new Error(`${name} is not a basic attribute`);
 		}
 
-		console.log(type);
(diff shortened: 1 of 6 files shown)


src/renderer/corporate/member/corporate-member-container.jsx
```diff
@@ -153,15 +153,19 @@ class CorporateMemberContainerComponent extends PureComponent {
...
+		const type =
+			this.state.type === 'individual'
+				? individualAttributeTypes[name]
+				: corporateAttributeTypes[name];
 
 		if (!type || !type.content) {
 			throw new Error(`${name} is not a basic attribute`);
 		}
 
+		console.log(type);
```
sk91: stray log

I usually prepend "XXX" to debug logs, that way I can do a quick search on "XXX" before submitting and removing them

PR #1943, opened 2019-12-11
Feat/1919 bank accounts flow

sk91:


src/renderer/marketplace/bank-accounts/checkout/payment-complete.jsx
```diff
@@ -94,11 +95,25 @@ export const BankAccountsPaymentComplete = withStyles(styles)(props => {
...
+									and{' '}
+									<strong>
+										will contact you shortly on the e-mail you provided{' '}
+									</strong>
+									, to continue the process. If you have any questions in the
+									meantime, you can reach us at:
+								</Typography>
+							)}
+
+							{!simpleFlow && (
```
andregoncalves: Works for now. 
But I this is a very good example of where your idea of composable UI flows would be used.

PR #1956, opened 2019-12-17
fix: issue with categories array index when loading selfkey data

andregoncalves:
Fixes issue loading data from selfkey airtable.
Occured if more than one data airtable exists.


andregoncalves:
Don't merge, fixing tests

src/main/marketplace/inventory/inventory-service.js
```diff
@@ -64,12 +64,14 @@ export class InventoryFetcher {
...
 		let inventory = await this.fetchInventory();
-		const categories = inventory
+		const itemsCategory = inventory
 			.filter(itm => itm.status === 'active')
 			.map(itm => itm.category);
 
+		const categories = [...new Set(itemsCategory)];
+
 		const data = await Promise.all(
-			[...new Set(categories)].map(async category => {
```
sk91: the set is used to ensure array is unique, check out this change does not break anything

andregoncalves: Yes, it was moved above that line because it's used again later:
`const categories = [...new Set(itemsCategory)];`

Issue #2031, opened 2020-02-06
Optional KYC requirements bug


andregoncalves:
Optional required documents should be shown but not block submission if empty

rhdelima:
@andregoncalves Failing this one. When editing optional requirement, an error is encountered. See recording: http://recordit.co/6qOLuV7T5N

rhdelima:
Looks good. Pass

PR #2040, opened 2020-02-12
fix: 2031 - optional kyc requirements are no longer mandatory

andregoncalves:
#2031


src/renderer/kyc/application/current-application-container.jsx
```diff
@@ -46,11 +46,14 @@ class CurrentApplicationComponent extends PureComponent {
 		}
 		const error = requirements.reduce((acc, curr) => {
 			if (acc) return acc;
-			if (!curr.options || !curr.options.length) return true;
+			if (!curr.required) return false;
```
sk91: should this be applied only to empty attributes?

andregoncalves: You are correct!
Pushed an update

Issue #2054, opened 2020-02-18
Transactions Screen Fixes


oanamangiurea:
- [x]  Table filters should both be aligned to the left https://gyazo.com/e01db9dc6c8ee4b283f84ebdb69a7625 

- [x] Box area is not standard, it's missing a border line and rounded corners

- [x] Table items should also be left aligned, even on empty table https://gyazo.com/39938c047160134e066da47b403f93d7

- [x] Follow the same style for the no transactions yet text, like we use for other tables. Ex: https://projects.invisionapp.com/d/main#/console/15222689/386105859/preview 

- [x]  Use our own styling for dropdown https://gyazo.com/4dca3fe6b95b02613129eefe775b907e

rhdelima:
Hi @oanamangiurea ,
Let me know if you're good with the fix. If you want to check on your end, please use env variable below to show transaction filters: FEATURE_TRANSACTIONS_LIST_FILTER=1


![image](https://user-images.githubusercontent.com/10294014/76739540-4b5ff280-67a7-11ea-805b-dbd7d2fc1e05.png)

![image](https://user-images.githubusercontent.com/10294014/76739640-734f5600-67a7-11ea-9aab-068419a1c7d2.png)

oanamangiurea:
@rhdelima almost. Dropdown for transactions type should have the following options:
"Send", "Receive", "Buy", "Swap" (uppercase first letter)
And the date in the date-picker input, should be the following format: DD-MM-YYYY, as it's displayed in the table below
@designhorf FYI

rhdelima:
@oanamangiurea Let me know if you're good with this now. However, I would like to raise the following issues:
1. Filter doesn't work after selecting type and/or date
2. Based on design, date field should accept range. 

![image](https://user-images.githubusercontent.com/10294014/77630269-b4085580-6f85-11ea-992f-2d3001f0be7d.png)

oanamangiurea:
@rhdelima not ok with this, especially if filter doesn't work after selecting something. And we should also have the date range, I think the component is already in the storybook, time to use it.

designhorf:
@oanamangiurea @rhdelima there is a separate ticket for filters https://github.com/SelfKeyFoundation/Identity-Wallet/issues/1979. 
This task fixed only layout alignment issues, please see the first comment  https://github.com/SelfKeyFoundation/Identity-Wallet/issues/2054#issue-566765828.

oanamangiurea:
[...]

PR #2065, opened 2020-02-21
Fix/1979 transactions list

designhorf:
Changes:
https://github.com/SelfKeyFoundation/Identity-Wallet/issues/2054


designhorf:
@sk91 good idea, I'll add feature flag

PR #2066, opened 2020-02-21
Feat/2036 custom token modal

sk91:


sk91:
@designhorf @oanamangiurea  send/receive transaction works. 
there is a small issue thoug, if there is a small amount of tokens/eth, it would show 0 as the amount, which is obviously wrong.

I fixed it to display the full amount, but the layout is not a bit broken
Screenshot 2020-03-02 at 18 13 34

Another thing that I found frustrating it that there is no easy way to copy wallet address now.

I need to go inside send/receive, click on "receive" tab and only then i can copy it. was it done intentionally?

oanamangiurea:
@sk91 yes, the decision to hide the copy of the address button was done intentionally, as during user testing, the users never used it from the dashboard, they were "blind" to that button, they were opening either KEY or ETH popups to copy it from there. 

@designhorf  if the amount doesn't fit with the token name on a single line, then it should fall on a new one, something like: [Screenshot from Gyazo](https://gyazo.com/35a7760e6a002105325235b668d9bd2f)

designhorf:
@oanamangiurea okay, I'm changing it

sk91:
@designhorf also fixed number of decimal places in crypto-manager table

PR #2114, opened 2020-03-25
feat: Loans Marketplace container and list page

andregoncalves:


andregoncalves:
A few notes:
- Range component is not included, will be after new selfkey-ui update is merged.
- 30 day average column is not included (not available yet in airtable)

PR #2149, opened 2020-03-29
fix: alignment of ledger table

designhorf:
Changes:
- Fixed alignment of address table
Screenshot 2020-03-29 at 15 45 04


src/renderer/wallet/unlock/select-address.jsx
```diff
@@ -143,6 +150,7 @@ class SelectAddress extends PureComponent {
 				justify="flex-start"
 				alignItems="flex-start"
 				spacing={40}
+				className="lofaszKa"
```
andregoncalves: small thing: this class "lofaszKa" doesn't seem to be used anywhere.

PR #2169, opened 2020-04-14
feat: totle swap use contract address

andregoncalves:
Use contract address for token swap.
Force KEY to use a hard coded address to avoid swapping for the "wrong" key token.


src/renderer/transaction/swap/index.jsx
```diff
@@ -94,6 +94,8 @@ const styles = theme => ({
 	}
 });
 
+const KEY_ADDRESS = '0x4cc19356f2d37338b9802aa8e8fc58b0373296e7';
```
sk91, line 97: This looks a bit ugly, if you don't want to deal with it now, maybe add a todo to think about it later at least

andregoncalves: Refactored.

Issue #2198, opened 2020-04-29
Loans Listing+Details UI Fixes


oanamangiurea:
1.  Change icon for calculator to the one from design. It’s already in the storybook https://gyazo.com/ad7246d39e4de167a3f28f312e471cd5
2.  Loan Calculator box https://gyazo.com/b3347919e4911b20d859de1ada5f7ae2:
    1. The box should be dismissible, in design I added a X icon to close it, and we need to remember that the user closed it 
    2. Box padding should be:  30px 24px 60px; 
    3. Icon padding should be: 0 24px 24px; 
3. Table filters: https://gyazo.com/52c49d423bf4e32529f28718b2cbc4fc 
    1. Button group should look like a group, not like 2 buttons touching. I think the component is already in the storybook. The 2nd button should be “Centralized” not “Licensed”
    2. Current rates slider is missing the values for the range, plus it should be much wider, so we can fit those values
    3. Assets accepted dropdown it’s missing it’s placeholder “Choose…”
4. Table Listing: https://gyazo.com/629cbf67f5072213ec242aae5fe2e575
    1. Table cell it’s missing top/bottom padding (issue raised as well for the material ui upgrade task)
5. Details View: https://gyazo.com/7b36ded49afe7fcc4a0d967e21762fc8
    1. Title of the box, the grey text next to the loan provider should state the platform type, not say “Loans”
    2. Padding of the entire box is too big on the right, should have same padding bot left/right
    3. Button is too small, it’s container should be MuiGrid-grid-xs-4 from xs-3
    4. There’s an extra horizontal line before the tabs, it shouldn’t be there
    5. Assets should be separated by a comma

rhdelima:
@oanamangiurea @andregoncalves Original issues were good on my end. However, I saw few more. Let me know if need to open a separate ticket.

1. When opening Loans Marketplace, P2P is selected by default but listing shows ALL
![image](https://user-images.githubusercontent.com/10294014/80966310-dde74e80-8e46-11ea-94b5-06b6a3825a95.png)

2. Overlapped texts on Current Rates slider. See some screenshots below.
![image](https://user-images.githubusercontent.com/10294014/80966359-fa838680-8e46-11ea-8641-d5b28e8846e7.png)
![image](https://user-images.githubusercontent.com/10294014/80966698-93b29d00-8e47-11ea-98e0-cab8d131276a.png)

andregoncalves:
Fix on PR review
[...]

PR #2251, opened 2020-05-18
2198 support multiple fiat currencies

andregoncalves:
Addresses #2198 - displaying loan data in different currencies.
Gets fiat currency exchanges rates when needed from api.exchangeratesapi.io free service.
Allows loan calculations to be displayed in USD, EUR and GBP.


andregoncalves:
Please Review #2246 first.

PR #2273, opened 2020-05-21
2254 load tokens to wallet

sk91:


andregoncalves:
@sk91 can you check failing unit tests? (src/main/token/token.spec.js)

sk91:
tests fixed

PR #2384, opened 2020-09-09
2364 notarization terms update

andregoncalves:
- New wallet TOS
- Removed notarization specific terms popup (it's now covered in general TOS)


sk91 (approved):
LGTM

PR #2389, opened 2020-09-14
feat(inventory): continue sync after single upsert failure

sk91:
if one of the synced inventory items fails, sync other items anyway


andregoncalves (approved):
LGTM

PR #2402, opened 2020-09-28
fix: remove link if empty in airtable

andregoncalves:
Do not render links if terms or privacy for vendor in airtable are empty

Screenshot 2020-09-28 at 10 05 48


sk91 (approved):
LGTM

PR #2443, opened 2020-11-12
fix: prevent error transactions from continously syncing

andregoncalves:
Prevent error transactions from continuously resyncing with blockchain.


sk91 (approved):
LGTM

PR #2469, opened 2020-12-01
fix: show existing DID if already created

andregoncalves:
For users that already created a DID, show it in the selfkey profile.


sk91 (changes requested):
We should not do this. The did is not used anywhere...

The applications with did that we've seen is coming from mobile where did was not properly disabled

andregoncalves:
We're not enabling DID, only showing it for people that have created it before.
There were some inquiries on support about it. 
I don't see why this would be a problem as DID is not used directly by the user.

sk91:
Oki, as long and we understand that it is not used anywhere


Max

sk91 (approved):

PR #2470, opened 2020-12-01
Refactor/remove unused selfkey profile components

andregoncalves:
Removed and refactor unused old selfkey profile components and folders


sk91 (approved):
LGTM

PR #2476, opened 2020-12-08
feat: create hd wallet

sk91:
use hd wallet flow for create wallet, instead of keystore

### DIFF
--- a/src/renderer/wallet/create/password/password-util.js
+++ b/src/renderer/wallet/create/password/password-util.js
@@ -11,6 +11,7 @@ export const handlePassword = (event, state) => {
 		4: 'Strong'
 	};
 	const newPassword = event.target.value;
+	console.log(newPassword);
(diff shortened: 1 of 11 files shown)


src/renderer/wallet/create/password/password-util.js
```diff
@@ -11,6 +11,7 @@ export const handlePassword = (event, state) => {
 		4: 'Strong'
 	};
 	const newPassword = event.target.value;
+	console.log(newPassword);
```
andregoncalves, line 14: stray log

andregoncalves (approved):
Looks good. Great work!

src/renderer/wallet/create/password/password-util.js
sk91: will remove in a different pull request, almost ready

PR #2480, opened 2020-12-08
feat: disallow submissions from restricted jurisdictions

andregoncalves:
Screenshot 2020-12-08 at 11 35 33


sk91 (approved):

sk91 (approved):
LGTM,

But the user can still come to the page with no attributes, and fill the info on kyc-checklist popup

andregoncalves:
> LGTM,
> 
> But the user can still come to the page with no attributes, and fill the info on kyc-checklist popup

We handle those cases, check current-application-container and popup.

PR #2481, opened 2020-12-08
Wallet connect

sk91:


maycon-mello (commented):
LGTM

src/main/application/deeplink-service.js
```diff
@@ -0,0 +1,57 @@
...
+			if (!match || match.length !== 3) {
+				log.info('invalid deepl ink format');
+				return;
+			}
+			if (!this.handlers[match[1]]) {
+				log.info('no handler for %s', match[1]);
+				return;
+			}
+
+			await this.handlers[match[1]].handleUrlCommand(match[2]);
```
andregoncalves, line 35: This is causing issues.
We receive the following url:
`selfkey://wc?uri=wc:b18792a6-2b02-4479-83ab-d8dd4d2ef9c6@1?bridge=https%3A%2F%2Fbridge.walletconnect.org&key=c28407a557a89e96be04a0fc44f1ad9a1c4c180d46af6ec27dfb03d917674089
`
which doesn't match that URL regex.

PR #2487, opened 2020-12-10
fix: hover state to listing elements

designhorf:
Changes:

- hover state to listing elements
Screenshot 2020-12-10 at 06 22 49


andregoncalves (approved):
Looks good.

PR #2539, opened 2020-12-30
2530 Save wallet

andregoncalves:
Asks if users want to create password and save wallet when unlocking existing wallet via PK or HD.

Screenshot 2020-12-30 at 10 24 13


maycon-mello (approved):
LGTM

